PR #17, opened 2021-11-18
[master, executor]: refine the implemement of heartbeats

hanfei1991:
This PR

1. Set timeout for every rpc request, including heartbeat.
2. The master's ttl should be executors ttl + rpc timeout
3. The master has a init ttl before the first heartbeat arrive.
4. We shouldn't send tasks to executor before it sends the first heartbeat.
5. refine the config.


executor/server.go
```diff
@@ -106,33 +107,37 @@ func (s *Server) Start(ctx context.Context) error {
...
+				Ttl:        uint64(s.cfg.KeepAliveTTL.Milliseconds() + s.cfg.RPCTimeout.Milliseconds()),
 			}
-			// FIXME: set timeout.
-			resp, err := s.cli.SendHeartbeat(ctx, req)
+			resp, err := s.cli.SendHeartbeat(ctx, req, s.cfg.RPCTimeout)
 			if err != nil {
 				log.L().Error("heartbeat meet error", zap.Error(err))
-				if s.lastHearbeatTime.Add(time.Duration(s.cfg.KeepAliveTTL) * time.Millisecond).Before(time.Now()) {
+				if s.lastHearbeatTime.Add(s.cfg.KeepAliveTTL).Before(time.Now()) {
 					return err
```
amyangfei, line 141: Should we return `heartbeat timeout` or wrapping `err` with `heartbeat time`

hanfei1991: Yeah, that's a good point!

executor/server.go
```diff
@@ -106,42 +107,46 @@ func (s *Server) Start(ctx context.Context) error {
...
-	ticker := time.NewTicker(time.Duration(s.cfg.KeepAliveInterval) * time.Millisecond)
+	ticker := time.NewTicker(s.cfg.KeepAliveInterval)
+	s.lastHearbeatTime = time.Now()
 	for {
 		select {
 		case <-ctx.Done():
 			return nil
 		case <-exitCh:
 			return nil
 		case t := <-ticker.C:
```
amyangfei, line 126: In L148 `s.lastHearbeatTime` is set to t, is it better to set `s.lastHearbeatTime = time.Now()`, since heartbeat rpc could cost some time.

hanfei1991: We aim to keep `lastHbTime of executor` consistant with `lastHbTime of Master`. If we set the heartbeat time of executor to the `start time` of rpc, it will a little bit ealier than the heartbeat time of master, which is safe. In contrast, if we set it to the `end time` of rpc, it might be a little bit later than master's, which might cause that master wait for less time than executor. This gap is unsafe.

amyangfei (approved):

executor/config.go
```diff
@@ -20,9 +20,9 @@ import (
 	"flag"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/BurntSushi/toml"
-	"github.com/pingcap/failpoint"
```
liuzix, line 26:
```suggestion
```

liuzix (approved):
rest LGTM

hanfei1991:
This PR

- Set timeout for every rpc request, including heartbeat.
- The master's ttl should be executors ttl + rpc timeout
- The master has a init ttl before the first heartbeat arrive.
- We shouldn't send tasks to executor before it sends the first heartbeat.
- refine the config.

PR #18, opened 2021-11-19
makefile: add goimports to tools bin

amyangfei:
Fix make error from root directory

```
./generate-proto.sh: line 74: goimports: command not found
make: *** [Makefile:4: df-proto] Error 127
```


hanfei1991 (approved):
LGTM, that's cool!

PR #19, opened 2021-11-21
jobmaster/demo: add rate limit in benchmark master

amyangfei:
- Add rate limit to task rescheduling when error happens
- Typo fix in readme

### DIFF
--- a/master/jobmaster/benchmark/master.go
+++ b/master/jobmaster/benchmark/master.go
@@ -3,6 +3,7 @@ package benchmark
 import (
 	"context"
 	"sync"
+	"time"
 
 	"github.com/hanfei1991/microcosom/master/cluster"
 	"github.com/hanfei1991/microcosom/model"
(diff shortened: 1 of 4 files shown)


master/jobmaster/benchmark/master.go
```diff
@@ -226,10 +256,23 @@ func (m *Master) monitorSchedulingTasks() {
 			// cancel it
 			//}
 
-			log.L().Logger.Info("begin to reschedule task group")
+			log.L().Logger.Info("begin to reschedule task group", zap.Any("group", group))
 			if err := m.reScheduleTask(group); err != nil {
 				log.L().Logger.Error("cant reschedule task", zap.Error(err))
-				// FIXME: this will cause deadlock problem
```
hanfei1991, line 232: I think this problem still remain when we read and write the same channel `scheduleWaitingTasks` and this channel is full. So leave this comments and we should address it in the furture

hanfei1991: but after we schedule "a group of tasks", this problem is mitigated

amyangfei: Oh yes, the channel could be full, I misunderstand the meaning.

hanfei1991 (approved):

PR #20, opened 2021-11-21
[test] introduce mock-grpc module for integration tests

hanfei1991:
use mock-grpc to test communication between master and executors


test/heartbeat_test.go
```diff
@@ -0,0 +1,93 @@
+package test_test
```
amyangfei, line 1: why is this package `test_test`?

executor/master-client.go
```diff
@@ -2,10 +2,12 @@ package executor
 
 import (
 	"context"
-	"errors"
 	"time"
 
 	"github.com/hanfei1991/microcosom/pb"
+	"github.com/hanfei1991/microcosom/pkg/errors"
```
amyangfei, line 8: the import path should be updated

test/heartbeat_test.go
hanfei1991: indeed this is an integration-test. Which package do you think is better?

executor/master-client.go
hanfei1991: shouldn't it be "github.com/hanfei1991/microcosm/pkg/errors" ?

test/heartbeat_test.go
amyangfei: Oh I see, the test is put in a separated package.

amyangfei (approved):

PR #21, opened 2021-11-22
Replace some pkgs with dm pkgs

amyangfei:
- replace pkg/log with pingcap/ticdc/dm/pkg/log
- replace pkg/etcdutil with pingcap/ticdc/dm/pkg/etcdutil
- remove pkg/helper
- remove pkg/retry


hanfei1991:
The `go.sum` is so spectacular. Is it expected?

amyangfei:
> The `go.sum` is so spectacular. Is it expected?

yep, I have ran `go mod tidy`, it does have so many changes.

hanfei1991 (approved):

PR #22, opened 2021-11-22
.*: use ticdc rfc errors, to keep the same style with tidb

amyangfei:
Use the rfc errors defined in TiDB. ref: https://github.com/pingcap/tidb/blob/master/docs/design/2020-05-08-standardize-error-codes-and-messages.md


hanfei1991 (approved):
Great Job! LGTM

PR #30, opened 2021-11-25
*: add a global flow-id field as single dataflow identifier

amyangfei:
- A little refactor, add a `flow-id` field to distinguish user submit jobs
- sink output to different dir based on the `flow-id`


hanfei1991:
Is flow-id to distinguish users?

amyangfei:
> Is flow-id to distinguish users?

not exactly, it acts like `DM task id` or `cdc changefeed name`, maybe we can also add a struct to contain more information of dataflow, such as 
```go
type Auto struct {
    FlowID string
    UserID int64
    // ...
}
```

amyangfei:
I don't know why circleCI is not running.

hanfei1991:
I just set circle ci applied to all forks. I don't know why it doesn't work

hanfei1991 (approved):

PR #31, opened 2021-11-26
jobmaster: move system level jobmaster code outside demo

amyangfei:
- Refactor the code structure of job master.
- Move some system level code outside demo.

### DIFF
--- a/master/jobmaster/jobmanager.go
+++ b/master/jobmaster/jobmanager.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/hanfei1991/microcosm/master/cluster"
 	"github.com/hanfei1991/microcosm/master/jobmaster/benchmark"
+	"github.com/hanfei1991/microcosm/master/jobmaster/system"
 	"github.com/hanfei1991/microcosm/model"
 	"github.com/hanfei1991/microcosm/pb"
 	"github.com/hanfei1991/microcosm/pkg/autoid"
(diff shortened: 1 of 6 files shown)


master/jobmaster/jobmanager.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/hanfei1991/microcosm/master/cluster"
 	"github.com/hanfei1991/microcosm/master/jobmaster/benchmark"
+	"github.com/hanfei1991/microcosm/master/jobmaster/system"
```
nongfushanquan, line 9: jobmanager.go should be under the  master  folder ,since it is a part of server master ,not the job master

master/jobmaster/jobmanager.go
```diff
@@ -17,7 +18,7 @@ import (
 // JobManager manages all the job masters, and notify the offline executor to them.
 type JobManager struct {
 	mu         sync.Mutex
-	jobMasters map[model.JobID]JobMaster
+	jobMasters map[model.JobID]system.JobMaster
 
 	idAllocater    *autoid.Allocator
 	resourceMgr    cluster.ResourceMgr
```
nongfushanquan, line 24: job manager  should not hold the resource manger

amyangfei: yes, I will refactor it in another PR.

master/jobmaster/jobmanager.go
amyangfei: addressed

amyangfei:
ptal @nongfushanquan @hanfei1991

hanfei1991 (approved):

PR #32, opened 2021-11-29
[test]: make demo-test mimic dm workload.

hanfei1991:
## Refine Demo

In this PR, every subplan of demo job is composed of 

1. **Receive op** receives data for mock grpc server, like a mysql binlog that contains multiple tables. The record could be `DATA` or `DDL`
2. **Syncer op** (unfinished) processes DDL records. It communicates with `JobMaster` and wait until all other syncers receive the same DDL records.
3. **Sink op** pass the `DATA` record to a testContext. We will check if any tests miss or occur repeatly.

## Refine runtime

A task can read from multiple inputs. In every `poll` action, a task will :
1. Ask operator which input should be read. operator will return an index indicating the input should be read. If the index is -1, any input is ok.
2. We read a chunk of data from input channel, but push the records one by one. If the operator is blocked, we will push the rest of data to `input cache`.


executor/master-client.go
```diff
@@ -73,3 +73,7 @@ func (c *MasterClient) RegisterExecutor(ctx context.Context, req *pb.RegisterExe
 	defer cancel()
 	return c.client.RegisterExecutor(ctx1, req)
 }
+
+func (c *MasterClient) SubmitJob(ctx context.Context, req *pb.SubmitJobRequest) (resp *pb.SubmitJobResponse, err error) {
+	return c.client.SubmitJob(ctx, req)
```
nongfushanquan, line 78: why should executor submit job?

hanfei1991: that's `master-client` under executor directory. I will move this file out to avoid misunderstand.

amyangfei (approved):

executor/runtime/task.go
```diff
@@ -58,29 +59,39 @@ func (c *Channel) writeBatch(records []*Record) ([]*Record, bool) {
...
-	inputs []*Channel
-	output []*Channel
-	//	waker  func()
-	ctx *taskContext
+	cfg         *model.Task
+	id          model.TaskID
+	status      int32
+	inputCache  []Chunk
+	outputCache []Chunk
+	op          operator
```
nongfushanquan, line 86: only one operator in a task

PR #34, opened 2021-11-30
[runtime]: put operator buidling logic into different folders

hanfei1991:
We use a `OpBuilder` interface to anaylze operator config bytes and return a operator object.
The procedure is:

- Register builder logic to `OpBuilders`
- When runtime generates operators, it gets `OpBuilder` from the global map and build the wanted operator.
- The OpBuilder is implemented in different pkgs. For now we only have `benchmark`. We are supposed to support DM and CDC in same manner.


executor/runtime/benchmark/operator.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (o *opSyncer) syncDDL(ctx *runtime.TaskContext) {
+	time.Sleep(1 * time.Second)
+	ctx.Wake()
+}
+
+func (o *opSyncer) Next(ctx *runtime.TaskContext, r *runtime.Record, _ int) ([]runtime.Chunk, bool, error) {
+	record := r.Payload.(*pb.Record)
+	if record.Tp == pb.Record_DDL {
+		go o.syncDDL(ctx)
```
amyangfei, line 156: Is there any place to wait sync DDL finished.

hanfei1991: When sync DDL finished, it will call `ctx.Wake()` function. If it is failed, I havn't figured out a good way to notify, which will be considered in the future.

hanfei1991: and the `Wake` function will reset the task status and push it back to running queue.

amyangfei (approved):

PR #35, opened 2021-12-02
[demo]: refine demo logic

hanfei1991:
This PR adapts the test kit `producer` to tasks and run it in executor runtime. The changes include
## Customized JobMaster

I change the `JobMaster` interface to `DispatchTasks(ctx, tasks)` from `DispatchJob(ctx)`, which allow us to dispatch tasks incrementally. 
I also add an interface called `Start()` which is implemented by different workloads. For example, `benchmark` needs to start producer first and start `recv-sync-sink` second. This change allow us to dispatch different stages of tasks.

## Refine runtime

In poll function, operator implements `NextWantedIndexIdx` and return two special marks

- **DontNeedData** For operator which reads data from (mock) grpc, they don't need data from task inputs.
- **DontRequireIdx** This operator want any data as fast as possible and don't care about whitch input are the data from. For example, SQL operator `Union` perfectly fit in this sort.


hanfei1991:
TODO: The use of this demo has been changed, I will update the README files.

executor/runtime/benchmark/operator.go
```diff
@@ -181,3 +188,152 @@ func (o *opSink) Next(ctx *runtime.TaskContext, r *runtime.Record, _ int) ([]run
...
+}
+
+func (o *opBinlog) NextWantedInputIdx() int {
+	if o.cacheRecord != nil {
+		return runtime.DontNeedData
+	}
+	return runtime.DontRequireIndex
+}
+
+func (o *opBinlog) Next(ctx *runtime.TaskContext, r *runtime.Record, _ int) ([]runtime.Chunk, bool, error) {
```
amyangfei, line 298: What is the return `[]runtime.Chunk` used for? seems it is always `nil`

amyangfei: Oh, is it caused by interface definition, not all operators return chunk in `Next`

hanfei1991: yep. the design of interface has to be compatible

amyangfei:
rest LGTM

executor/runtime/benchmark/operator.go
```diff
@@ -181,3 +188,152 @@ func (o *opSink) Next(ctx *runtime.TaskContext, r *runtime.Record, _ int) ([]run
...
+		if o.pk >= o.dataCnt {
+			return outputData, true, nil
+		}
+		o.pk++
+		if o.pk == 10000 {
+			log.L().Info("got 10000 data")
+		}
+		if o.pk%o.ddlFrequency == 0 {
+			o.schemaVer++
+			ddlLock.Lock()
```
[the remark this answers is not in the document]
hanfei1991: Acctually No.  I wanted to keep ddl in order for every `binlog` channel, but I should find a cleaner way. I will fix it

amyangfei (approved):

PR #40, opened 2021-12-06
metastore: add metastore server and simple query API

### LINKED ISSUE #38: [server,executor] add basic metadata abstraction
TODO list
- [ ] Add basic metadata abstraction used for failover, sample test case
   - Start one master, one executor
   - Restart master
   - The executor should recover, but met error `[ERROR] [server.go:231] ["heartbeat meet error"] [error="[DFLOW:ErrUnknownExecutorID]cannot find executor ID: 1"]`

amyangfei:
This PR introduces a new metastore server as a meta service in dataflow engine, close #38

## metastore service
It is an independent service, manages multiple metastore and handles RPC requests from metastore user. The abilities it provides including:                                             
- A **service discovery metastore** management, it is am embed etcd indeed.    
- A metastore for dataflow engine **system** , it shares the same etcd in service discovery metastore, and we will separate them later. We share the same etcd just for prototype verification.
- A metastore for applications running on dataflow engine, which is not implemented currently.

### API
The metastore service provides two APIs current
- register API, meta store can register itself to metastore service
- query API, dataflow system, application can query to metastore service for available metastore address.

## MetaKV API
Use the same function signature as etcd API, to support KV related operations.


hanfei1991:
In my mind, HA store should be an abstract client which connects different backend, instead of a service that wraps backend.
Is it necessary to introduce a service layer? If the meta writting is frequent, will the sole HA server become a bottleneck?

amyangfei:
> In my mind, HA store should be an abstract client which connects different backend, instead of a service that wraps backend.
> Is it necessary to introduce a service layer? If the meta writting is frequent, will the sole HA server become a bottleneck?

Yes, correct. In fact this server should not provide meta store, instead it only records the available backend meta store and tells system, or application the metastore address when they are querying. The embed etcd is used in service discovery, while the metastore used by dataflow engine system will be separated later

hanfei1991:
> > In my mind, HA store should be an abstract client which connects different backend, instead of a service that wraps backend.
> > Is it necessary to introduce a service layer? If the meta writting is frequent, will the sole HA server become a bottleneck?
> 
> Yes, correct. In fact this server should not provide meta store, instead it only records the available backend meta store and tells system, or application the metastore address when they are querying. The embed etcd is used in service discovery, while the metastore used by dataflow engine system will be separated later

Alright!

amyangfei:
This meta server works as the manager of backend meta stores

hanfei1991:
I agree with the sole role of meta store manager, but is it worthy to make it a seperate server? How about integrating it to master-server?

pkg/metadata/key.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"strings"
+
+	"github.com/hanfei1991/microcosm/pkg/errors"
+	"github.com/pingcap/ticdc/dm/pkg/log"
+	"go.uber.org/zap"
+)
+
+const (
+	// MetaStoreKeyBase is the common prefix of keys in dataflow system
```
hanfei1991, line 12: I have added a `pkg/adapter` for the key prefix management. It is ported by DM. What do you think if we should resue that or just manage them here?

amyangfei:
> I agree with the sole role of meta store manager, but is it worthy to make it a seperate server? How about integrating it to master-server?

Yep, that is a good suggestion, I removed the separated metastore server and integrated it into server master.

hanfei1991 (approved):
LGTM

Issue #4, opened 2021-10-18
[Design of Master] How to implement Resource Manager (and Resource Model).


hanfei1991:
The Resource Manager is in charge of recording the resource consumption and alloc the free resource for job master.

## Resource Model

I think the biggest problem for now is how to measure the usage of resource. There are three dimension of resource in a SQL workload, that is 

- cpu
- memory
- i/o

I have no idea whether it fits in the cdc/dm workload.

nongfushanquan:
ResourceManager should alway be in running mode, receiving the executor's workload information , accepting the request to allocate executor for new task.

PR #44, opened 2021-12-13
[resource manager]: add separated resouce manager interface

### LINKED ISSUE #16: Resource allocation should be implemented by the ResourceManager
https://github.com/hanfei1991/mircocosom/blob/f9f0de8e04d79c1fe2c3567c24b5b80c4524f889/master/jobmaster/benchmark/master.go#L122

JobMaster focus on how to create subjob and monitor task status etc.

amyangfei:
A refactor for #4, also close #16

- Add a standalone resource package, and a `RescMgr` interface.
- The executor manager interacts with `RescMgr` as follows
   - The executor manager calls `Register` and `Update` API provided by `RescMgr` when new executor is registered, or status/usage update.
   - The executor manager asks `RescMgr` for resource
   - The `RescMgr` allocates executors to tasks

After this refactor, server master doesn't need to interact with resource directly, instead it works as follows

server master <-> executor manager <-> resource manager


master/resource/manager.go
```diff
@@ -0,0 +1,30 @@
+package resource
+
+import (
+	"github.com/hanfei1991/microcosm/model"
+	"github.com/hanfei1991/microcosm/pb"
+)
+
+// RescMgrr manages the resources of the clusters.
```
amyangfei, line 8:
```suggestion
// RescMgr manages the resources of the clusters.
```

master/resource/capacity_impl.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"sync"
+
+	"github.com/hanfei1991/microcosm/model"
+	"github.com/hanfei1991/microcosm/pb"
+	"github.com/hanfei1991/microcosm/pkg/errors"
+)
+
+// CapRescMgr implements ResourceMgr interface, and it uses node capacitiy as
```
hanfei1991, line 11:
```suggestion
// CapRescMgr implements ResourceMgr interface, and it uses node capacity as
```

master/resource/capacity_impl.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+// Allocate implements RescMgr.Allocate
+func (m *CapRescMgr) Allocate(tasks []*pb.ScheduleTask) (bool, *pb.TaskSchedulerResponse) {
+	return m.allocateTasksWithNaiveStrategy(tasks)
+}
+
+// Update implements RescMgr.Update
+func (m *CapRescMgr) Update(id model.ExecutorID, use RescUnit, status model.ExecutorStatus) error {
```
hanfei1991, line 40: When an executor is put offline, we should also remove it from resource manager.

master/resource/capacity_impl.go
amyangfei: addressed

master/resource/capacity_impl.go
amyangfei: Good catch, addressed

master/jobmaster/system/impl.go
```diff
@@ -22,9 +22,8 @@ type Master struct {
 	ctx    context.Context
 	cancel func()
 
-	resourceManager cluster.ResourceMgr
-	client          cluster.ExecutorClient
-	mClient         cluster.JobMasterClient
+	client  cluster.ExecutorClient
+	mClient cluster.JobMasterClient
```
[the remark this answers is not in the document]
amyangfei: This change is caused by gofmt, and is not relevant to this PR. I will refactor job master in another PR.

amyangfei:
PTAL @hanfei1991

hanfei1991 (approved):

PR #46, opened 2021-12-15
client: decouple clients out of executor managers

hanfei1991:
Since we plan to seperate the job master, we must maintain executor/master clients in a single object. 

- This PR decouple the client and executor manager.
- Add address info to schedule result, so that job master can know the address of executors.


client/executor_client.go
```diff
@@ -26,11 +26,11 @@ type executorClient struct {
 	client pb.ExecutorClient
 }
 
-func (c *executorClient) close() error {
-	return c.conn.Close()
-}
+//func (c *executorClient) close() error {
+//	return c.conn.Close()
+//}
```
amyangfei, line 31: Can we remove these lines

master/cluster/executor_manager.go
```diff
@@ -220,22 +208,22 @@ func (e *ExecutorManager) checkAliveImpl() error {
...
-		exec.mu.Lock()
-		if exec.Status == model.Running {
-			exec.Status = model.Disconnected
-		}
-		exec.mu.Unlock()
-		return resp, err
-	}
-	return resp, nil
-}
+//func (e *ExecutorManager) Send(ctx context.Context, id model.ExecutorID, req *ExecutorRequest) (*ExecutorResponse, error) {
```
amyangfei, line 211: ditto

hanfei1991: done

amyangfei (approved):

Issue #42, opened 2021-12-09
Make server master failover can be tested


amyangfei:
After https://github.com/hanfei1991/microcosm/pull/41 we have supported to deploy multiple server maters for HA, we still need some features to make executor can work when server master suffers leader transfer, including 
- [x] Support gRPC redirect to leader in server master #56
- [x] Support server master to load metadata to rebuild executors information when it changes to leader OR the executor maintains a list of server master and updates leader periodically	#47 
- [x] Support server master to rebuild job master information.

nongfushanquan:
Load metadata to restore the job manager

amyangfei:
One failover scenario:
- Server master, one job with a job master and several tasks are running.
- Server master crashes and restarts, it reloads enough information.
- Then cancel job.

Questions:
1. Should server master cancel job master, and job master cancels other tasks. Or server master cancels all tasks belonging to this job?
2. If server master cancels all tasks, it should persist metadata of each task.
3. Who is responsible for task rescheduling, job master? If so, should server master know rescheduling?

hanfei1991:
1. Server master manages job masters, so it's ok to cancel job master. However the rest tasks should be stopped by job master. You can refer to the current implemention just as what I write.
2. I think server master can see all tasks in the meta store and job master is a special kind of task. Server master should load and reschedule all the job masters, and each job master manages its own tasks.
3. Reschedule happens when executor/tasks crash, and job master should be in charge of rescheduling. However, resource manager should answer job master the schedule result.

amyangfei:
more tasks ref: #190

PR #47, opened 2021-12-16
master: support single master failover

amyangfei:
Part of #42 

This PR has several changes to the server master.

- Introduce an `initialized` field, before server loads information from metastore and initializes, this field is `false` and some server gRPC API will return `MasterNotReady` error.
- Support auto recovery of executor states in server master.


master/server.go
```diff
@@ -166,6 +198,31 @@ func (s *Server) Start(ctx context.Context) (err error) {
 	if test.GlobalTestFlag {
 		return s.startForTest(ctx)
 	}
+
```
nongfushanquan, line 202: ToDo : ServerMaster should update the its address in etcd , so the executor can get it

master/server.go
```diff
@@ -166,6 +198,31 @@ func (s *Server) Start(ctx context.Context) (err error) {
...
+	// rebuild states from existing meta if needed
+	err = s.resetExecutor(ctx)
+	if err != nil {
+		return err
+	}
+
+	// start background managers
+	s.executorManager.Start(ctx)
+	s.jobManager.Start(ctx)
```
nongfushanquan, line 220: the failover of jobmanager will be  implemented in the JobManager Start()  method ?

amyangfei: Will be implemented in another PR.

master/server.go
amyangfei: It will be implemented in another PR.

nongfushanquan (approved):

PR #48, opened 2021-12-16
[jobmaster]: immigrate job master from master to executor

hanfei1991:
## Move JobMaster

This PR move job master to executor. The idea is:

- treat job master as a normal task to be dispatched. For adaption, this PR
  - Start job master in `operator prepare` phase.
  - Monitor job master in `operator poll` phase.
  - Cancel task in `operator close` phase.
- This PR also adjust `jobmanager` to a "meta jobmaster" that manages all running jobmaster. **Every job master is a normal task in Job Manager** This PR
  - Dispatch job master by dispatching tasks
  - Stop job master by stoping tasks.

## Uniform Job ID and Task ID

As job master is a kind of task, we use int64 for both task id and job id. In the code, they are all unified as `model.ID`.
There difference is:

- Job ID is a int32 number.
- Task ID = (Job ID << 32) + int32 auto-id


hanfei1991:
Right now the structure of directories doesn't make sense any more. For instance, the jobmaster has no relationship with master any longer. I will adjust the structure in next PR

master/jobmanager.go
```diff
@@ -15,109 +15,100 @@ import (
...
-		err = errors.ErrBuildJobFailed.GenWithStack("unknown job type", req.Tp)
+		err := errors.ErrBuildJobFailed.GenWithStack("unknown job type", req.Tp)
 		resp.Err = errors.ToPBError(err)
 		return resp
 	}
 	log.L().Logger.Info("finished build job")
 	j.mu.Lock()
 	defer j.mu.Unlock()
-	err = jobMaster.Start(ctx)
+	log.L().Logger.Info("finished dispatch job")
```
amyangfei, line 93: Does this log need to be after L 99?

master/jobmanager.go
```diff
@@ -15,109 +15,100 @@ import (
...
-func (j *JobManager) startImpl(ctx context.Context) {
-	for {
-		select {
-		case execID := <-j.offExecutors:
-			log.L().Logger.Info("notify to offline exec")
-			j.mu.Lock()
-			for _, jobMaster := range j.jobMasters {
-				jobMaster.OfflineExecutor(execID)
-			}
-			j.mu.Unlock()
```
amyangfei, line 44: Will the offline logic be added in another PR?

proto/master.proto
```diff
@@ -72,12 +72,12 @@ message SubmitJobRequest {
...
 
 message CancelJobRequest {
-    int32 job_id = 1;
+    int64 job_id = 1;
 }
 
 message SubmitJobResponse {
     Error err = 1;
-    int32 job_id = 2;
+    int64 job_id = 2;
```
amyangfei, line 80: Where do we restrict the job_id is an `int32`

master/jobmanager.go
hanfei1991: I think so, I should be added after we make sure the logic of heartbeats

amyangfei (approved):

PR #49, opened 2021-12-16
remove the jobmaster to the parent floder,and refactor the jobmaster …

nongfushanquan:
This PR has some changes about the job master

- move the job master folder to the parent's folder since the job master will not run in the servermaster. 
- redefine the jobmaster interface


hanfei1991:
good job but master has been updated since #48 is merged ~

PR #50, opened 2021-12-17
remove the jobmaster floder to the parent floder,and add the jobmaste…

nongfushanquan:
This PR has some changes about the job master

move the job master folder to the parent's folder since the job master will not run in the servermaster.
redefine the jobmaster interface


jobmaster/jobmaster.go
```diff
@@ -0,0 +1,120 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package JobMaster
```
hanfei1991, line 14: What is this for?

PR #53, opened 2021-12-20
master: support leader to serve only and return leader info

amyangfei:
1. Add leader election in server master
2. Check whether the server master is leader when searving gPRC API, if not returns NotLeader error and contains leader information.
3. Add a update members goroutine in server master


master/server.go
```diff
@@ -264,15 +242,98 @@ func (s *Server) startGrpcSrv() (err error) {
...
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-s.session.Done():
+		return errors.ErrMasterSessionDone.GenWithStackByArgs()
+	}
+}
+
+func (s *Server) apiPreCheck() *pb.Error {
```
nongfushanquan, line 302: if not the leader, the request will be  forwarded in this method ?

nongfushanquan (approved):

master/server.go
amyangfei: The forward feature will be added later, it doesn't affect the main workflow. Another feature will be added is the gRPC client can find not leader and auto selection to real leader.

PR #55, opened 2021-12-21
master: implements Election interface

liuzix:
- We need a mockable and testable object for leader elections.

### DIFF
--- a/master/campaign.go
+++ b/master/campaign.go
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	"github.com/hanfei1991/microcosm/pkg/errors"
-	"github.com/pingcap/log"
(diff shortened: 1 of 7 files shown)


master/campaign.go
```diff
@@ -2,57 +2,38 @@ package master
 
 import (
 	"context"
-	"time"
 
 	"github.com/hanfei1991/microcosm/pkg/errors"
-	"github.com/pingcap/tiflow/dm/pkg/log"
-	"go.etcd.io/etcd/mvcc"
+	"github.com/pingcap/log"
```
nongfushanquan, line 7: we can use the same log system in our project, including other files.

amyangfei (approved):

PR #58, opened 2021-12-22
persist all the job and tasks info

hanfei1991:
This pr writes job and tasks info to meta kv, and tries to recover from meta kv when starting. The details of persistant is:

- For every job, we write two records:
  - write the job config as <JobPrefix/JobID - Config>
  - write the job master running info as <TaskPrefix/JobID - TaskInfo>
  - during recovering, the job master can get the task ids from the task info of job master (yet unimplemented)
- For every task, we write one record:
  - write the task running info as <TaskPrefix/JobID - TaskInfo>


jobmaster/system/interface.go
```diff
@@ -4,18 +4,17 @@ import (
...
 )
 
 // JobMaster maintains and manages the submitted job.
 type JobMaster interface {
 	// DispatchJob dispatches new tasks.
-	DispatchTasks(ctx context.Context, tasks []*model.Task) error
+	DispatchTasks(tasks ...*model.Task)
 	// Start the job master.
-	Start(ctx context.Context) error
+	Start(ctx context.Context, metaKV metadata.MetaKV) error
```
nongfushanquan, line 15: the meta server / message server should be inited when creating the jobmaster /jobmanager/task.

executor/runtime/jobmaster/operator.go
```diff
@@ -5,16 +5,22 @@ import (
...
 	master system.JobMaster
 }
 
-func (j *jobMasterAgent) Prepare() error {
-	return j.master.Start(context.Background())
+func (j *jobMasterAgent) Prepare(ctx *runtime.TaskContext) error {
+	if test.GlobalTestFlag {
+		j.metaKV = ctx.TestCtx.GetMetaKV()
+	}
+	return j.master.Start(context.Background(), j.metaKV)
```
amyangfei, line 23: Seems the context can't be canceled, maybe we can leave a TODO

amyangfei (approved):

executor/runtime/jobmaster/build.go
```diff
@@ -3,13 +3,25 @@ package jobmaster
...
 	"github.com/hanfei1991/microcosm/client"
 	"github.com/hanfei1991/microcosm/executor/runtime"
 	"github.com/hanfei1991/microcosm/jobmaster/benchmark"
 	"github.com/hanfei1991/microcosm/jobmaster/system"
 	"github.com/hanfei1991/microcosm/model"
+	"github.com/hanfei1991/microcosm/pb"
 	"github.com/hanfei1991/microcosm/pkg/errors"
+	"github.com/hanfei1991/microcosm/pkg/metadata"
+	"github.com/hanfei1991/microcosm/test"
+	"github.com/pingcap/tiflow/dm/pkg/log"
```
nongfushanquan, line 18: can we should use the same log system? in another file , use the PingCAP/log

executor/runtime/jobmaster/build.go
```diff
@@ -18,6 +30,44 @@ func RegisterBuilder() {
 
 type jobMasterBuilder struct{}
 
+func getEtcdMetaKV(ctx context.Context, clients *client.Manager) (metadata.MetaKV, error) {
```
nongfushanquan, line 33: why we need expose etcd here,  metaserver interface is not enough?

executor/runtime/jobmaster/build.go
```diff
@@ -18,6 +30,44 @@ func RegisterBuilder() {
...
+			}),
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+	return metadata.NewMetaEtcd(etcdCli), nil
+}
+
 func (b *jobMasterBuilder) Build(op model.Operator) (runtime.Operator, bool, error) {
```
nongfushanquan, line 71: what's the purpose of this file ?  if used to help create jobmaster, this file should be put under the jobmaster floder

jobmaster/system/impl.go
```diff
@@ -36,68 +39,88 @@ type Master struct {
...
 		scheduleWaitingTasks: make(chan scheduleGroup, 1024),
 		scheduleRateLimit:    rate.NewLimiter(rate.Every(time.Second), 1),
 
-		execTasks:    make(map[model.ExecutorID][]*model.Task),
-		runningTasks: make(map[model.ID]*Task),
+		runningTasks: make(map[model.ID]*model.Task),
 	}
 }
 
 func (m *Master) Cancel() {
```
nongfushanquan, line 56: what's the relationship between the master  and jobmaster

jobmaster/system/impl.go
```diff
@@ -36,68 +39,88 @@ type Master struct {
...
+		return err
+	}
+	m.mu.Lock()
+	m.runningTasks[taskInfo.ID] = taskInfo
+	// And wait task to report heartbeat.
+	m.mu.Unlock()
+	return nil
+}
+
+func (m *Master) updateEtcd(ctx context.Context, tasks []*model.Task) error {
```
nongfushanquan, line 93: why use the etcd directly instead of the megastore  here?

jobmaster/system/impl.go
```diff
@@ -250,40 +226,36 @@ func (m *Master) StopTasks(ctx context.Context, tasks []*model.Task) error {
...
 	// TODO: keep the goroutines alive.
-	go m.monitorExecutorOffline()
 	go m.monitorSchedulingTasks()
 }
 
+func (m *Master) addScheduleTasks(group scheduleGroup) {
+	if len(group) == 0 {
+		return
+	}
+	go func() {
```
nongfushanquan, line 242: why need create a  goroutie

master/jobmanager.go
```diff
@@ -24,18 +28,53 @@ type JobManager struct {
...
+		if err != nil {
+			return err
+		}
+		jobTask := &model.Task{
+			ID:   model.ID(jobID),
+			OpTp: model.JobMasterType,
+			Op:   kv.Value,
+			Cost: 1,
+		}
+		err = j.RestoreTask(ctx, jobTask)
```
nongfushanquan, line 53: why the jobmanager should need restore task?

hanfei1991: job master is a kind of task in job manager

jobmaster/system/impl.go
hanfei1991: to avoid the channel being blocked

jobmaster/system/impl.go
hanfei1991: yeah, the function name is wrong

jobmaster/system/impl.go
hanfei1991: "Master" is the base class of "JobMaster"

executor/runtime/jobmaster/build.go
hanfei1991: I have not figured out a good approach to get meta store from outside world, so here I get etcd address from master.

jobmaster/system/interface.go
hanfei1991: theoretically it should be. I will leave a comment

executor/runtime/jobmaster/build.go
hanfei1991: This file is set to analyze the incomming task/jobmaster config and build a runtime operator that drived by task.

PR #65, opened 2021-12-23
pkg/p2p: add some helper structs for integrating p2p

liuzix:
- Added some adapter logic to make `pkg/p2p` in `tiflow` easier to be used and mocked in Dataflow Engine.

### DIFF
--- a/pkg/p2p/message_handler_manager.go
+++ b/pkg/p2p/message_handler_manager.go
@@ -1,0 +1,149 @@
+package p2p
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"go.uber.org/atomic"
+	"go.uber.org/zap"
+)
+
+const (
+	defaultHandlerOperationTimeout = 3 * time.Second
+)
+
+// MessageHandlerManager is for managing message topic handlers.
+// NOTE: for each topic, only one handler is allowed.
+type MessageHandlerManager interface {
+	RegisterHandler(ctx context.Context, topic Topic, tpi TypeInformation, fn HandlerFunc) (bool, error)
+
+	UnregisterHandler(ctx context.Context, topic Topic) (bool, error)
+
+	CheckError(ctx context.Context) error
+
+	// Clean unregisters all existing handlers.
+	Clean(ctx context.Context) error
+
+	// SetTimeout sets the timeout for handler operations.
+	// A timeout is needed because the underlying handler operations are
+	// asynchronous in the MessageServer.
+	SetTimeout(timeout time.Duration)
+}
+
+func newMessageHandlerManager(registrar handlerRegistrar) MessageHandlerManager {
+	return &messageHandlerManagerImpl{
+		messageServer: registrar,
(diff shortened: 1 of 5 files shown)


pkg/p2p/message_handler_manager.go
```diff
@@ -0,0 +1,149 @@
...
+
+	CheckError(ctx context.Context) error
+
+	// Clean unregisters all existing handlers.
+	Clean(ctx context.Context) error
+
+	// SetTimeout sets the timeout for handler operations.
+	// A timeout is needed because the underlying handler operations are
+	// asynchronous in the MessageServer.
+	SetTimeout(timeout time.Duration)
```
hanfei1991, line 33: If we expose this function, does it mean we have demand of setting timeout on the fly?

pkg/p2p/message_handler_manager.go
```diff
@@ -0,0 +1,149 @@
...
+func (m *messageHandlerManagerImpl) RegisterHandler(
+	ctx context.Context,
+	topic Topic,
+	tpi TypeInformation,
+	fn HandlerFunc,
+) (bool, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if _, ok := m.topics[topic]; ok {
```
hanfei1991, line 71: Will it possible to replace the topic, or do we have to remove it at first before we replace it?

pkg/p2p/server.go
```diff
@@ -0,0 +1,117 @@
...
+
+	wg.Go(func() (err error) {
+		defer log.L().ErrorFilterContextCanceled("message server exited", zap.Error(err))
+		return errors.Trace(s.messageServer.Run(ctx))
+	})
+
+	wg.Go(func() (err error) {
+		defer func() {
+			// TODO (zixiong) filter out expected harmless errors.
+			log.L().Debug("grpc server exited", zap.Error(err))
```
hanfei1991, line 95: Will it always be nil if you don't assign it explictly?

pkg/p2p/message_handler_manager.go
liuzix: - Allowing setting the timeout in the constructor will make the constructor function too complicated.
- Making the timeout a constant will make unit testing difficult.
Although we do not have demand for resetting the timeout on the fly, I think exposing a method is the better way.

pkg/p2p/message_handler_manager.go
liuzix: We have to remove it as far as `MessageServer` is concerned. The implementation of the `MessageServer` does not support replacing it without removing it first. Can you think of a situation where exposing a `Replace` method is helpful?

pkg/p2p/server.go
liuzix: No. The deferred function can see the return value.

pkg/p2p/message_handler_manager.go
hanfei1991: No, I can't figure out this senario either. That's fine

pkg/p2p/message_handler_manager.go
hanfei1991: ok let it be

hanfei1991 (approved):

PR #68, opened 2021-12-24
server master: refine example config, add a join master sample

amyangfei:
Add a join master config sample, add scheme to `peer-urls` in server master sample config.


hanfei1991 (approved):

hanfei1991:
Maybe we should create a "sample" directory in the future, and put these example files together.

PR #76, opened 2022-01-04
executor: add task resource unit abstraction

### LINKED ISSUE #64: resource manager: support executor workload report
- executor should collect workload information, including how many tasks are running, CPU load etc.
- executor reports workload to resource manager. (how to report, together in heartbeat? or separate message channel)
- resource manager handles workload information.

amyangfei:
Close #64 

- Add a `TaskRescUnit` interface, which is used to pass resource usage from task(application) to executor.
- Executor collects resource usage periodically and reports to server master via gRPC API. (It can migrate to p2p message service in the future)


proto/master.proto
```diff
@@ -120,3 +124,17 @@ message ScheduleResult {
 message TaskSchedulerResponse {
     map<int64, ScheduleResult> schedule = 1;
 }
+
+message ExecWorkload {
+    JobType tp = 1;
```
nongfushanquan, line 129: maybe more content should be reported to the resource manager

nongfushanquan (approved):

proto/master.proto
amyangfei: yes, we can add when we need more fields.

Issue #6, opened 2021-10-25
[to-discuss] How should executors know the address of the master's leader.


hanfei1991:
In TiCDC, all the comminucation are done by etcd, so that's not a problem. In the new **p2p** framework, `MassageRouter` watch the etcd to apply the owner address change.
In DM, non-leader will transfer the request to leader.
In tikv-pd, there is a background thread to update the leader address.

nongfushanquan:
maybe we can follow the dm's method

amyangfei:
This issue has been separated into two parts
- executor needs to know the address change of server master. <- this is what this issue focus
- executor needs to know the address change of other executors. <- focused in https://github.com/hanfei1991/microcosm/issues/73

PR #77, opened 2022-01-05
client: support multi targets and round robin in master client

amyangfei:
Part job of https://github.com/hanfei1991/microcosm/issues/6

This PR supports multiple targets in server master gPRC client. Client can try each next server master if the current one failed.

In later PR more failover mechanism will be added, including 
- Auto recognition for server members.
- Server master leader specific request.


client/master_client.go
```diff
@@ -13,38 +16,64 @@ import (
...
+	log.L().Logger.Info("dialing master", zap.Strings("urls", c.urls))
+	for _, addr := range c.urls {
+		ctx, cancel := context.WithTimeout(ctx, dialTimeout)
+		conn, err := grpc.DialContext(ctx, addr, grpc.WithInsecure(), grpc.WithBlock())
+		if err != nil {
+			log.L().Warn("dial to one of server master failed", zap.String("addr", addr))
+			cancel()
+			continue
+		}
+		cancel()
```
nongfushanquan, line 42: why invoke the cancel function here?

amyangfei: This cancels the timeout context in L35, which is only used when dialing to server master.

nongfushanquan (approved):

PR #78, opened 2022-01-06
client: support auto member update in master client

amyangfei:
Part job of #6 
- Fix data race in master client
- Support to add more target conns in master client
- Add member update in HeartbeatResponse from server master to executor.
- Executor add conn in master client if needed.


client/master_client.go
```diff
@@ -24,25 +25,51 @@ type clientHolder struct {
...
+// master that is not maintained in current MasterClient.
+func (c *MasterClient) UpdateClients(ctx context.Context, urls []string) {
+	for _, addr := range urls {
+		// TODO: refine address with and without scheme
+		addr = strings.Replace(addr, "http://", "", 1)
+		if _, ok := c.clients[addr]; !ok {
+			c.urls = append(c.urls, addr)
+			log.L().Info("add new server master client", zap.String("addr", addr))
+			err := c.dialMaster(ctx, addr)
+			if err != nil {
```
nongfushanquan, line 58: if error , still store the address in the c.urls?

amyangfei: Good catch, added some refine code and unit test.

amyangfei:
PTAL @nongfushanquan

master/member.go
```diff
@@ -28,34 +28,28 @@ func (m *Member) Unmarshal(data []byte) error {
...
-		if isServLeader && (!exist || currLeader.Name != m.Name) {
-			s.leader.Store(&Member{
-				Name:         m.Name,
-				IsServLeader: true,
-				IsEtcdLeader: true,
-				Addrs:        m.ClientURLs,
-			})
-			s.createLeaderClient(ctx, m.ClientURLs)
-		}
+		isEtcdLeader := m.ID == etcdLeaderID
```
nongfushanquan, line 42: what's the relationship between the server leader and. etcd leader ?

amyangfei: etcd leader is elected first, which is based the raft election in embed etcd.
server leader is based on etcd leader, but it could have short latency, and we can observe 
- etcd leader, server leader in the same node
- etcd leader exists, server leader doesn't exist for a short time.

nongfushanquan (approved):

Issue #60, opened 2021-12-23
server master: add API to support job operation


amyangfei:
Job operations include:
- submit job
- cancel job
- query job status

Current we have implemented submit/cancel job via server gRPC API.

amyangfei:
Besides we should consider multi-tenant for job storage.

amyangfei:
support query job in #171 
support pause job in #228

PR #79, opened 2022-01-06
*: support suspend jobs, part 1

hanfei1991:
part job of #60 
This PR supports the suspending of jobs and tasks, but it hasn't considered the consistency after fail-over.

The "suspend" is implemented in an async way, because suspending might happen when rescheduling or fail recovering.
After the user sends suspend request, we mark the "target status" for the target tasks. A goroutine will check the status regularly, and will try to suspend it util success. There are some work to do furture.

1. The commands should be serializable. If several commands are sent in the same time, we should order them.
2. The command have to be persisted and to be continued when node crashes.


client/executor_client.go
```diff
@@ -34,6 +34,8 @@ func (c *executorClient) Send(ctx context.Context, req *ExecutorRequest) (*Execu
 		resp.Resp, err = c.client.SubmitBatchTasks(ctx, req.SubmitBatchTasks())
 	case CmdCancelBatchTasks:
 		resp.Resp, err = c.client.CancelBatchTasks(ctx, req.CancelBatchTasks())
+	case CmdSuspendBatchTasks:
+		resp.Resp, err = c.client.SuspendBatchTasks(ctx, req.SuspendBatchTasks())
```
amyangfei, line 38: What about using `pause` and `resume` task, which keeps the same style in DM and TiCDC.

hanfei1991: Good suggestion, I will change it later

amyangfei (approved):

PR #81, opened 2022-01-10
jobmaster: refine check running tasks in job master

amyangfei:
Before this PR we check running tasks with a 300ms rate limiter, but it will bump too rate limit log every 300ms.
- Change it to a 1s ticker.
- Use free port in unit test instead of hard code ports.

### DIFF
--- a/jobmaster/system/impl.go
+++ b/jobmaster/system/impl.go
@@ -48,7 +48,6 @@ type Master struct {
 	scheduleWaitingTasks chan scheduleGroup
 	// rate limit for rescheduling when error happens
 	scheduleRateLimit *rate.Limiter
-	monitorTasksLimit *rate.Limiter
 }
 
 type TaskStatus int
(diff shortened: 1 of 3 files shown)


jobmaster/system/impl.go
```diff
@@ -310,39 +308,33 @@ func (m *Master) addScheduleTasks(group scheduleGroup) {
...
 		select {
 		case <-m.ctx.Done():
 			return
-		default:
+		case <-timer.C:
+			m.checkRunningTasks()
 		}
-		delay := m.monitorTasksLimit.Reserve().Delay()
-		if delay != 0 {
-			log.L().Logger.Warn("monitor task rate limit", zap.Duration("delay", delay))
```
hanfei1991, line 321: I copied this log from reschedule logic. How about just removing it

amyangfei: Seems a ticker or rate limiter has the same effect.

hanfei1991: Yeah, that's true.

hanfei1991 (approved):

PR #82, opened 2022-01-11
master: use advertise-addr in server master membership

amyangfei:
- Add a membership interface in server master
- Use `AdvertiseAddr` in member struct, which is more accurate for server discovery.

### DIFF
--- a/master/member.go
+++ b/master/member.go
@@ -4,17 +4,20 @@ import (
 	"context"
 	"encoding/json"
 
+	"github.com/hanfei1991/microcosm/pkg/adapter"
+	"github.com/hanfei1991/microcosm/pkg/errors"
 	"github.com/pingcap/tiflow/dm/pkg/log"
+	"go.etcd.io/etcd/clientv3"
 	"go.uber.org/zap"
 )
 
 // Member stores server member information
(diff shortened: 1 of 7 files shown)


master/member.go
```diff
@@ -28,25 +31,72 @@ func (m *Member) Unmarshal(data []byte) error {
...
+	GetConfigs(ctx context.Context) (map[string]*Config, error)
+	GetMembers(ctx context.Context, leader *Member, etcdLeaderID uint64) ([]*Member, error)
+}
+
+type EtcdMembership struct {
+	etcdCli *clientv3.Client
+}
+
+func (em *EtcdMembership) GetConfigs(ctx context.Context) (map[string]*Config, error) {
+	resp, err := em.etcdCli.Get(ctx, adapter.MasterInfoKey.Path(), clientv3.WithPrefix())
```
nongfushanquan, line 45: if the membership is a struct ,why not set  _the adapter.MasterInfoKey.Path()__  as a input parameter for this method

amyangfei: `MasterInfoKey` and `master.Config` are bounded logic here, so this function doesn't pass a key as parameter.

nongfushanquan (approved):

Issue #91, opened 2022-01-14
Integrate `MessageRouter` with cluster membership management mechanism.


amyangfei:
This issue contains two tasks
- [x] Maintain service discovery information in etcd, such as mapping from uuid to address. ref: https://github.com/hanfei1991/microcosm/pull/92
- [x] Add self-address maintenance mechanism in executor. #95 
- [x] Poll the mapping list, and call `AddPeer` or `RemovePeer` according to membership update. #96

PR #96, opened 2022-01-17
executor: add service discovery in executor server

amyangfei:
Finish task-3 in #91
- Rename `keepalive` to `discoveryKeepalive` in executor server, and refactor this function.
   - When meeting session done or watch error from discovery service, doesn't return an error and terminate executor, retry to connect to discovery service(meta store) instead. We do this because the aliveness of executor is decided by the **heartbeat loop**, the discovery keepalive loop is only used to maintain executor address information.
- Add unit test to `discoveryKeepalive`
- Integrate `MessageRouter` interface in executor server.


executor/server.go
```diff
@@ -50,6 +62,7 @@ func NewServer(cfg *Config, ctx *test.Context) *Server {
 		testCtx:     ctx,
 		cliUpdateCh: make(chan []string),
 	}
+	s.discoveryConnector = s.connectToEtcdDiscovery
```
hanfei1991, line 65: Why delegate this function instead of using it directly?

amyangfei: Use a function variable here to make it easy for unit test.

hanfei1991 (approved):

executor/server.go
```diff
@@ -131,6 +144,15 @@ func (s *Server) Stop() {
 		}
 	}
 
+	if s.metastore != nil {
```
liuzix, line 147: Can you add a comment why it is necessary to call `Delete` here?

executor/server.go
```diff
@@ -260,29 +282,96 @@ func (s *Server) keepalive(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+	executors, err := s.discovery.Snapshot(ctx)
 	if err != nil {
 		return err
 	}
+	for uuid, exec := range executors {
+		if uuid == string(s.info.ID) {
+			continue
```
liuzix, line 338: I think we should not skip adding itself to the router.

pkg/srvdiscovery/interface.go
```diff
@@ -17,18 +17,23 @@ type ServiceResource = model.ExecutorInfo
...
 
 	// Watch watches the change of service resources, the watched events will be
 	// returned through a channel.
 	Watch(ctx context.Context) <-chan WatchResp
+
+	// SnapshotClone returns the clone of cached snapshot in Discovery
+	SnapshotClone() (map[UUID]ServiceResource, Revision)
+
+	// CopySnapshot copies snapshot from given Discovery
+	CopySnapshot(snapshot map[UUID]ServiceResource, revision Revision)
```
liuzix, line 38: Probably we should comment on when `CopySnapshot` is useful?

executor/server.go
amyangfei: addressed

executor/server.go
amyangfei: added comment

pkg/srvdiscovery/interface.go
amyangfei: added comment

amyangfei:
PTAL @liuzix

liuzix (approved):

Issue #93, opened 2022-01-17
Modify `runtime` to provide semantics compatible to the master-worker framework


amyangfei:
- [x] Support calling `Poll` of worker periodically in runtime. #97
- [x] Implement `DispatchTask` rpc in executor with new worker model, which will be called from server master. #101
- [x] Implement worker register mechanism. #105

PR #97, opened 2022-01-17
executor/worker: new simple runtime

hanfei1991:
This PR implements a simple runtime for worker and implements:

- **Start(conn int)** means starting this runtime, with conn goroutines.
- **AddWorker(lib.Worker)** adds a worker to this runtime
- **Workerload() RescUsage** calculate the sum of running workloads.

I abstract the scheduler to implement more sophisticated algo in the future.
The next PR will connect them with outside world.

ref: #93

### DIFF
--- a/lib/fake/fake_worker.go
+++ b/lib/fake/fake_worker.go
@@ -1,0 +1,61 @@
+package fake
+
+import (
+	"context"
+	"errors"
+
+	"github.com/hanfei1991/microcosm/lib"
+	"github.com/hanfei1991/microcosm/model"
+	dcontext "github.com/hanfei1991/microcosm/pkg/context"
+)
+
+var _ lib.WorkerImpl = &dummyWorkerImpl{}
+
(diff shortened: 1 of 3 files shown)


lib/fake/fake_worker.go
```diff
@@ -0,0 +1,61 @@
...
+func (d *dummyWorkerImpl) Workload() (model.RescUnit, error) {
+	return model.RescUnit(10), nil
+}
+
+func (d *dummyWorkerImpl) OnMasterFailover(_ lib.MasterFailoverReason) error {
+	return nil
+}
+
+func (d *dummyWorkerImpl) CloseImpl() {
+	d.closed = true
```
amyangfei, line 56: Will `d.closed` cause data race with L33

hanfei1991: It will and fixed

amyangfei (approved):

Issue #94, opened 2022-01-17
Add unit tests for the implementation of new master-worker framework.


amyangfei:
- [x] https://github.com/hanfei1991/microcosm/pull/90
- [x] #98
- [x] #102 
- [x] #111

amyangfei:
- [x] https://github.com/hanfei1991/microcosm/pull/90

PR #98, opened 2022-01-18
lib/master: add unit test cases (Part I)

liuzix:
- Implemented mock dependencies for `BaseMaster`.
- Added two basic unit test cases for `BaseMaster`.
- Fixed some minor bugs.

### DIFF
--- a/executor/server.go
+++ b/executor/server.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hanfei1991/microcosm/executor/runtime"
 	"github.com/hanfei1991/microcosm/model"
 	"github.com/hanfei1991/microcosm/pb"
+	"github.com/hanfei1991/microcosm/pkg/config"
 	"github.com/hanfei1991/microcosm/pkg/errors"
 	"github.com/hanfei1991/microcosm/pkg/metadata"
 	"github.com/hanfei1991/microcosm/test"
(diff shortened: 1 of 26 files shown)


amyangfei:
ref #94

executor/server.go
```diff
@@ -32,7 +32,7 @@ type Server struct {
 
 	tcpServer   tcpserver.TCPServer
 	grpcSrv     *grpc.Server
-	cli         *client.MasterClient
+	cli         *client.MasterClientImpl
```
amyangfei, line 35: Can we use interface `MasterClient` here?

lib/master.go
```diff
@@ -215,6 +247,11 @@ func (m *BaseMaster) runWorkerCheck(ctx context.Context) error {
 }
 
 func (m *BaseMaster) OnError(err error) {
+	if errors.Cause(err) == context.Canceled {
+		// TODO think about how to gracefully handle cancellation here.
+		log.L().Error("BaseMaster is being canceled", zap.Error(err))
```
amyangfei, line 252: Is warning level better here?

lib/common.go
```diff
@@ -14,7 +14,9 @@ type (
 	WorkerStatusCode int32
 	WorkerType       int64
```
nongfushanquan, line 15: worker type should be defined as enum

liuzix: I will address this in another PR

nongfushanquan (approved):

PR #101, opened 2022-01-18
[executor]:support dispatch task

hanfei1991:
Support grpc interface to accepte dispatch worker


amyangfei:
ref task-2 in #93

lib/worker.go
```diff
@@ -79,6 +79,18 @@ func NewBaseWorker(
...
 }
 
+func (w *BaseWorker) ID() WorkerID {
+	return w.id
+}
+
+func (w *BaseWorker) Workload() model.RescUnit {
+	wl, err := w.impl.Workload()
+	if err != nil {
+		panic(err)
```
amyangfei, line 89: Better to use `log.L().Panic`.
Besides we may add some error passing mechanism in the feature, single worker error should not affect the whole runtime.

executor/server.go
```diff
@@ -200,6 +226,18 @@ func (s *Server) Run(ctx context.Context) error {
 		return nil
 	})
 
+	pollCon := s.cfg.PollConcurrency
+	if pollCon == 0 {
+		pollCon = sysruntime.NumCPU()
+	}
```
amyangfei, line 232: We can put this logic to `Config.Parse` function in executor/config.go

executor/server.go
```diff
@@ -129,8 +137,15 @@ func (s *Server) ResumeBatchTasks(ctx context.Context, req *pb.PauseBatchTasksRe
 	return &pb.PauseBatchTasksResponse{}, nil
 }
 
-func (s *Server) DispatchTask(ctx context.Context, request *pb.DispatchTaskRequest) (*pb.DispatchTaskResponse, error) {
-	panic("implement me")
+func (s *Server) DispatchTask(ctx context.Context, req *pb.DispatchTaskRequest) (*pb.DispatchTaskResponse, error) {
+	log.L().Info("dispatch task", zap.String("req", req.String()))
+	workerID := "dummy"
```
nongfushanquan, line 142: why  a fixed id?

amyangfei (approved):

lib/worker.go
```diff
@@ -83,6 +83,18 @@ func NewBaseWorker(
 	}
 }
 
+func (w *BaseWorker) ID() WorkerID {
+	return w.id
+}
+
+func (w *BaseWorker) Workload() model.RescUnit {
+	wl, err := w.impl.Workload()
```
amyangfei, line 91:
```suggestion
	wl, err := w.Impl.Workload()
```

PR #102, opened 2022-01-19
BaseMaster: Refine workerManager & add unit test cases

liuzix:
- Refined `workerManager`, which is a private component inside `BaseMaster`.
- Added unit test cases for `workerManager`.
- Fixed minor bugs.

ref #94


lib/worker_manager_test.go
```diff
@@ -0,0 +1,272 @@
...
+	"testing"
+	"time"
+
+	"github.com/benbjohnson/clock"
+	"github.com/gavv/monotime"
+	"github.com/hanfei1991/microcosm/pkg/p2p"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHeartBeatPingPongAfterCreateWorker(t *testing.T) {
```
amyangfei, line 14: Can these test cases run in parallel mode

lib/master.go
```diff
@@ -156,21 +157,6 @@ func (m *BaseMaster) Poll(ctx context.Context) error {
 		return errors.Trace(err)
 	}
 
-	for {
```
nongfushanquan, line 159: what's the purpose of those code?

nongfushanquan (approved):

lib/worker_manager_test.go
liuzix: Yes

amyangfei (approved):

PR #105, opened 2022-01-20
*: Implement worker registry

liuzix:
ref: task-3 in #93

- Implemented a registry for managing globally available workers.
- Minor adjustments to the implementation of `BaseWorker` and `Worker` interface.
- Adjusted affected UT of runtime.


executor/worker/runtime.go
```diff
@@ -109,15 +111,20 @@ func (r *Runtime) Start(conn int) {
...
+	r.workerList.Store(worker.WorkerID(), worker)
 	r.initingWorker <- worker
 }
 
 func (r *Runtime) Workload() model.RescUnit {
 	ret := model.RescUnit(0)
 	r.workerList.Range(func(_, value interface{}) bool {
 		worker := value.(lib.Worker)
-		ret += worker.Workload()
+		workload, err := worker.Workload()
```
nongfushanquan, line 122: no error returned for workload method is more reasonable

amyangfei (approved):

PR #107, opened 2022-01-20
registry: support passing WorkerConfig

liuzix:
- Registry now supports automatically unmarshalling worker configuration in `[]byte`

### DIFF
--- a/executor/server.go
+++ b/executor/server.go
@@ -155,7 +155,11 @@ func (s *Server) DispatchTask(ctx context.Context, req *pb.DispatchTaskRequest) 
 	}
 
 	newWorker, err := registry.GlobalWorkerRegistry().CreateWorker(
-		dctx, lib.WorkerType(req.TaskTypeId), lib.WorkerID(workerID), lib.MasterID(req.MasterId))
+		dctx,
+		lib.WorkerType(req.GetTaskTypeId()),
+		lib.WorkerID(workerID),
+		lib.MasterID(req.GetMasterId()),
+		req.GetTaskConfig())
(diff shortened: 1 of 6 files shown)


executor/server.go
```diff
@@ -155,7 +155,11 @@ func (s *Server) DispatchTask(ctx context.Context, req *pb.DispatchTaskRequest)
 	}
 
 	newWorker, err := registry.GlobalWorkerRegistry().CreateWorker(
-		dctx, lib.WorkerType(req.TaskTypeId), lib.WorkerID(workerID), lib.MasterID(req.MasterId))
+		dctx,
+		lib.WorkerType(req.GetTaskTypeId()),
+		lib.WorkerID(workerID),
+		lib.MasterID(req.GetMasterId()),
+		req.GetTaskConfig())
```
hanfei1991, line 162: Do we need to know "task type"?

lib/registry/factory.go
```diff
@@ -12,11 +16,39 @@ type WorkerFactory interface {
...
 	) (lib.Worker, error)
+
+	DeserializeConfig(configBytes []byte) (WorkerConfig, error)
+}
+
+type WorkerConstructor func(ctx *dcontext.Context, id lib.WorkerID, masterID lib.MasterID, config WorkerConfig) lib.Worker
+
+type SimpleWorkerFactory struct {
+	constructor WorkerConstructor
+	tpi         interface{}
```
amyangfei, line 29: Can have an explicit name, such as `configTpi`

amyangfei (approved):

PR #108, opened 2022-01-20
*: abstract some interfaces in server master, fix an executor panic

amyangfei:
- Abstract interface `ExecutorManager` and `JobManager` in server master
- Fix a panic in executor caused by nil executor info
- Add one unit test in etcd startup function in server master package.

### DIFF
--- a/executor/server.go
+++ b/executor/server.go
@@ -250,12 +250,12 @@ func (s *Server) Run(ctx context.Context) error {
 	s.workerRtm = worker.NewRuntime(ctx)
 	s.workerRtm.Start(pollCon)
 
-	err = s.startMsgService(ctx, wg)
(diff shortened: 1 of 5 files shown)


executor/server.go
```diff
@@ -254,12 +254,12 @@ func (s *Server) Run(ctx context.Context) error {
...
 	s.workerRtm.Start(pollCon)
 
-	err = s.startMsgService(ctx, wg)
+	err = s.selfRegister(ctx)
 	if err != nil {
 		return err
 	}
 
-	err = s.selfRegister(ctx)
+	err = s.startMsgService(ctx, wg)
```
amyangfei, line 262: Before self register, executor info is nil

nongfushanquan (approved):

PR #111, opened 2022-01-21
worker: add unit test cases

liuzix:
- Add unit test cases for `BaseWorker`.
- Abstracted out `pkg/clock` for mocking the clock.
- Fixed bugs in `BaseWorker`.

### DIFF
--- a/lib/worker.go
+++ b/lib/worker.go
@@ -5,8 +5,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/gavv/monotime"
 	"github.com/hanfei1991/microcosm/model"
+	"github.com/hanfei1991/microcosm/pkg/clock"
 	derror "github.com/hanfei1991/microcosm/pkg/errors"
 	"github.com/hanfei1991/microcosm/pkg/metadata"
 	"github.com/hanfei1991/microcosm/pkg/p2p"
@@ -48,16 +48,22 @@ type BaseWorker struct {
 	Impl WorkerImpl
 
 	messageHandlerManager p2p.MessageHandlerManager
-	messageRouter         p2p.MessageSender
+	messageSender         p2p.MessageSender
 	metaKVClient          metadata.MetaKV
 
 	masterClient *masterClient
+	masterID     MasterID
 
 	id            WorkerID
 	timeoutConfig TimeoutConfig
 
 	wg    sync.WaitGroup
 	errCh chan error
+
+	cancelMu      sync.Mutex
+	cancelBgTasks context.CancelFunc
+
+	clock clock.Clock
 }
 
 func NewBaseWorker(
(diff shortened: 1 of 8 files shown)


lib/worker.go
```diff
@@ -312,28 +368,31 @@ func (m *masterClient) HandleHeartbeat(msg *HeartbeatPongMessage) {
...
+	if msg.Epoch < m.masterEpoch {
+		log.L().Info("epoch does not match, ignore stale heartbeat",
 			zap.Any("msg", msg),
 			zap.Int64("master-epoch", m.masterEpoch))
 		return
 	}
+
+	if msg.Epoch > m.masterEpoch {
+		m.masterEpoch = msg.Epoch
+	}
```
amyangfei, line 380: Should `msg.Epoch == m.masterEpoch` be never happened.

lib/worker_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+	worker.On("InitImpl", mock.Anything).Return(nil)
+	worker.On("Status").Return(WorkerStatus{
+		Code: WorkerStatusNormal,
+	}, nil)
+	err := worker.Init(ctx)
+	require.NoError(t, err)
+
+	worker.clock.(*clock.Mock).Add(defaultTimeoutConfig.workerHeartbeatInterval + 1*time.Second)
+	worker.clock.(*clock.Mock).Add(defaultTimeoutConfig.workerHeartbeatInterval + 1*time.Second)
```
amyangfei, line 48: These two lines are the same, is it by design?

pkg/clock/clock.go
```diff
@@ -0,0 +1,60 @@
...
+
+type Clock interface {
+	After(d time.Duration) <-chan time.Time
+	AfterFunc(d time.Duration, f func()) *Timer
+	Now() time.Time
+	Since(t time.Time) time.Duration
+	Sleep(d time.Duration)
+	Tick(d time.Duration) <-chan time.Time
+	Ticker(d time.Duration) *Ticker
+	Timer(d time.Duration) *Timer
```
amyangfei, line 26: What about reusing the interface `clock.Clock` from github.com/benbjohnson/clock for these APIs.

lib/worker.go
```diff
@@ -147,7 +172,13 @@ func (w *BaseWorker) MetaKVClient() metadata.MetaKV {
 	return w.metaKVClient
 }
 
-func (w *BaseWorker) startBackgroundTasks(ctx context.Context) {
+func (w *BaseWorker) startBackgroundTasks() {
+	ctx, cancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
liuzix: Passing a context here will cause additional implicit assumptions about what context the `Executor` is passing us. We are expecting a long-lasting context, so I think using `context.Background()` is safer.

lib/worker.go
liuzix: This should be the norm.

lib/worker_test.go
liuzix: Yes.

liuzix: It seems that the mock clock's ticker needs this to trigger a tick.

amyangfei (approved):

Issue #112, opened 2022-01-21
Add k8s based deployment tools and e2e tests


amyangfei:
Currently this project can be tested manually, and lacks automatic integration tests. It is desired to add some development tools and e2e tests.
- [ ] Add docker-compose configs to support run demo cluster
- [ ] Add simple e2e tests based on docker-compose, the tests may be run with Github action.
- [ ] Support more complicated deployment with K8s maybe.

PR #115, opened 2022-01-24
[sample]: add dockerfile and dockercompose

hanfei1991:
relate to #112 

# Intro

This component shows how to start the cluster and run some tests by docker tools chain.

## Build

To build this repo, you can simply run `./prepare.sh`. Make sure you have installed docker and have run the docker daemon.

## Deploy

There are several configure files to use. The file name suggests the number of running server-master and executor nodes. For example, `1m1e.yaml` means this file contains one server-master and one executor. Use `./deploy.sh 1m1e.yaml` to deploy cluster.

## Destroy

Use `./stop.sh 1m1e.yaml` to destroy the cluster.

## Client

`client.sh` is used to send command to server-master. For old workload, we can use `./client.sh submit-job --master-addr server-master:10240 --config ./cmd/master-client/bench-example.toml` to run the simple workload.
Note that although we can successfully send the job, the job can't run as expected right now because the chaos of normal address and advertise address. We should manage to prove that the new worker-master framework can launch successfully.


cmd/master-client/main.go
```diff
@@ -33,6 +33,7 @@ func main() {
 	clt, err := client.NewMasterClient(ctx, []string{addr})
 	if err != nil {
 		fmt.Printf("err: %v", err)
+		os.Exit(0)
```
amyangfei, line 36:
```suggestion
		os.Exit(1)
```

liuzix (approved):

PR #116, opened 2022-01-24
job-manager: add new job manager with p2p message model

amyangfei:
- Add a new job manager v2, implemented based on lib/master.
- Start message server in server master.
- Use an independent `MasterMetaKey` in lib meta.


client/client_manager.go
```diff
@@ -10,7 +10,9 @@ import (
 )
 
 type ExecutorClientManager interface {
+	MasterClient() MasterClient
```
liuzix, line 13: Why is it necessary for an `ExecutorClientManager` to expose a `MasterClient`?

amyangfei: The interface name is not accurate, it is a client manager interface actually, which manages both executor clients and master client.

liuzix: Can we change the interface name?

amyangfei: Yep, changed it to `ClientsManager`

liuzix (approved):

PR #119, opened 2022-01-26
discovery: fix duplicated watch from discovery service

amyangfei:
- Fix watch from discovery multiple times.
- Reset watch channel after discovery service is reset.
- Close old watch channel when discovery service is reset.


executor/server.go
```diff
@@ -383,6 +385,15 @@ func (s *Server) createSession(ctx context.Context, etcdCli *clientv3.Client) (m
 	return session, nil
 }
 
+func (s *Server) resetDiscovery(ctx context.Context) (metaStoreSession, error) {
+	session, err := s.discoveryConnector(ctx)
```
hanfei1991, line 389: Now that `discoveryConnector` has replaced the `discovery` object. Why do we have to update discoveryWather instead of using `discovery.Watch` directly?

executor/server.go
```diff
@@ -359,6 +360,7 @@ func (s *Server) connectToEtcdDiscovery(ctx context.Context) (metaStoreSession,
 		etcdCli, adapter.ExecutorInfoKeyAdapter, defaultDiscoverTicker)
 	if old != nil {
 		s.discovery.CopySnapshot(old.SnapshotClone())
+		s.discovery.Close()
```
hanfei1991, line 363: I'm confused with here. Why close new discovery instead of close the old one?

amyangfei: addressed, it should close the old one.

executor/server.go
amyangfei: If we use `discovery.Watch` directly, it will create a watch channel each time when that case branch hits.

executor/server.go
hanfei1991: Then is there a unit-test covering this?

amyangfei: Not yet, this function is mocked in unit test.

amyangfei: the function `connectToEtcdDiscovery` will be refactored in next PR.

executor/server.go
hanfei1991: Got it...

hanfei1991: We'd better to add a comment above `Watch` function to warn the users.

executor/server.go
hanfei1991: it's fine

hanfei1991 (approved):

PR #123, opened 2022-01-26
resource: make heartbeats update reserved resource

hanfei1991:
`Reserve` means the expected max resource usage of this node.
`Used` means the resource actually used of this node.
after decoupling resource manager and job master, both of used and reserved resources should be updated by heartbeats.

### DIFF
--- a/servermaster/executor_manager.go
+++ b/servermaster/executor_manager.go
@@ -103,7 +103,8 @@ func (e *ExecutorManagerImpl) HandleHeartbeat(req *pb.HeartbeatRequest) (*pb.Hea
 	exec.heartbeatTTL = time.Duration(req.Ttl) * time.Millisecond
 	exec.Status = model.ExecutorStatus(req.Status)
 	usage := model.RescUnit(req.GetResourceUsage())
-	err := e.rescMgr.Update(exec.ID, usage, exec.Status)
+	// TODO: update reserve resources by heartbeats.
+	err := e.rescMgr.Update(exec.ID, usage, usage, exec.Status)
 	if err != nil {
 		return nil, err
 	}
--- a/servermaster/resource/capacity_impl.go
+++ b/servermaster/resource/capacity_impl.go
@@ -51,14 +51,15 @@ func (m *CapRescMgr) Allocate(tasks []*pb.ScheduleTask) (bool, *pb.TaskScheduler
 }
 
 // Update implements RescMgr.Update
-func (m *CapRescMgr) Update(id model.ExecutorID, use model.RescUnit, status model.ExecutorStatus) error {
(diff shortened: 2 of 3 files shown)


servermaster/resource/capacity_impl.go
```diff
@@ -51,14 +51,15 @@ func (m *CapRescMgr) Allocate(tasks []*pb.ScheduleTask) (bool, *pb.TaskScheduler
...
+func (m *CapRescMgr) Update(id model.ExecutorID, used, reserved model.RescUnit, status model.ExecutorStatus) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	exec, ok := m.executors[id]
 	if !ok {
 		return errors.ErrUnknownExecutorID.GenWithStackByArgs(id)
 	}
-	exec.Used = use
+	exec.Used = used
+	exec.Reserved = reserved
```
amyangfei, line 62: I forget the meaning of `Reserved`, would you please add some comments in the definition of `ExecutorResource`, about the meaning of `capacity`, `reserved`, `used` and how to use them. 😂

hanfei1991: ok

servermaster/executor_manager.go
```diff
@@ -103,7 +103,8 @@ func (e *ExecutorManagerImpl) HandleHeartbeat(req *pb.HeartbeatRequest) (*pb.Hea
 	exec.heartbeatTTL = time.Duration(req.Ttl) * time.Millisecond
 	exec.Status = model.ExecutorStatus(req.Status)
 	usage := model.RescUnit(req.GetResourceUsage())
-	err := e.rescMgr.Update(exec.ID, usage, exec.Status)
+	// TODO: update reserve resources by heartbeats.
+	err := e.rescMgr.Update(exec.ID, usage, usage, exec.Status)
```
amyangfei, line 107: Does `reserve` mean the resource manager allocates some resource (the resource can be either used or unused but prepare to be used)

hanfei1991: Right, the resource manager allows the tasks to use this amount of resource. The `Reserved` resource is estimated to be used, same as the `Cost` of tasks. If we use a container to run a task, we can limit the resource exactly by the `Reserved` amount.

hanfei1991: However, things always go sideway. In the go (or java like what spark runs on) runtime, the tasks can use much more (or less) resources than expected. In this case, there might be a global resource pool that allows the tasks to alloc more resource than the reserved. If the global resource pool is exhuasted, then oom occures.

nongfushanquan: why use the same value here

amyangfei (approved):

PR #128, opened 2022-01-28
master: fix worker offline detection

liuzix:
- Fix the problem where an executor is offline and the `OnWorkerOfflined` method is not called for affected masters.


amyangfei (approved):

lib/master.go
```diff
@@ -259,6 +250,15 @@ func (m *BaseMaster) runWorkerCheck(ctx context.Context) error {
 				return errors.Trace(err)
 			}
 		}
```
nongfushanquan, line 254: why change the order of those code

PR #134, opened 2022-02-06
[cmd/master-client]: refine the use of master clt

hanfei1991:
## Use

Right now, run demo task with command

```
./bin/master-client run-fake --master-addr 0.0.0.0:10240 --executor-addr "0.0.0.0:10241"
```

Note that `master-addr` is nessesary.

## Change

Use `cobra` to implement clt. This package organize commands orderly.


ctl/cmd.go
```diff
@@ -0,0 +1,62 @@
...
+		return err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), rpcTimeout)
+	defer cancel()
+
+	resp, err := cltManager.ExecutorClient(model.ExecutorID(execID)).Send(ctx, &client.ExecutorRequest{
+		Cmd: client.CmdDispatchTask,
+		Req: &pb.DispatchTaskRequest{
+			TaskTypeId: int64(lib.CvsJobMaster),
+			TaskConfig: []byte(`{"srcHost":"127.0.0.1:1234","srcDir":"data","dstHost":"127.0.0.1:1234","dstDir":"data2"}`),
```
nongfushanquan, line 51: replace the hard code  with the configure file

nongfushanquan (approved):

ctl/cmd.go
hanfei1991: Suppose to refine in the future.

PR #137, opened 2022-02-07
[WIP]Add etcd kv client

maxshuang:
Add etcd kv client implement for metastore client kv interface.

- Support Idempotent/TTL/Key Range/From Key/Key Prefix/Sort/Limit attributes
- Support prefix namespace isolation


pkg/metaclient/kv.go
```diff
@@ -0,0 +1,47 @@
...
+	// Any Op error will cause entire txn rollback and return error
+	Commit(ctx context.Context) (*TxnResponse, error)
+}
+
+type KV interface {
+	// Put puts a key-value pair into metastore.
+	// Note that key,value can be plain bytes array and string is
+	// an immutable representation of that bytes array.
+	// To get a string of bytes, do string([]byte{0x10, 0x20}).
+	// When passed WithTTL(ttl), Put will insert a key with ttl duration.
```
liuzix, line 19:
```suggestion
	// When WithTTL(ttl) is passed, Put will insert a key with ttl duration.
```
Please fix grammar in similar sentences.

pkg/metaclient/kv.go
```diff
@@ -0,0 +1,47 @@
...
+	// When passed WithRevision(revision), Put will be executed with corresponding revision, or do nothing on vice verse.
+	Put(ctx context.Context, key, val string, opts ...OpOption) (*PutResponse, error)
+
+	// Get retrieves keys.
+	// By default, Get will return the value for "key", if any.
+	// When passed WithRange(end), Get will return the keys in the range [key, end).
+	// When passed WithFromKey(), Get returns keys greater than or equal to key.
+	// When passed WithPrefix(), Get returns keys with prefix.
+	// When passed WithLimit(limit), the number of returned keys is bounded by limit.
+	// When passed WithSort(), the keys will be sorted.
```
liuzix, line 29: Please comment on the mutual exclusiveness of these options.

pkg/metaclient/kv.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+type KV interface {
+	// Put puts a key-value pair into metastore.
+	// Note that key,value can be plain bytes array and string is
+	// an immutable representation of that bytes array.
+	// To get a string of bytes, do string([]byte{0x10, 0x20}).
+	// When passed WithTTL(ttl), Put will insert a key with ttl duration.
+	// When passed WithRevision(revision), Put will be executed with corresponding revision, or do nothing on vice verse.
+	Put(ctx context.Context, key, val string, opts ...OpOption) (*PutResponse, error)
```
nongfushanquan, line 21: why the opts is needed?

pkg/metaclient/common.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Txn Response
+type TxnResponse struct {
+	Header *ResponseHeader
+	// Responses is a list of responses corresponding to the results from applying
+	// success if succeeded is true or failure if succeeded is false.
+	Responses []ResponseOp
+}
+
+type ResponseOp struct {
```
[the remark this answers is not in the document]
maxshuang: I think union is the most proper type for this ResponseOp. 
But it may seems strange in go for an intermediate  interface layer.

pkg/metaclient/kv.go
maxshuang: Get it.

pkg/metaclient/kv.go
maxshuang: Options carry some attributes in interface. In this case, options can carry the ttl and revision attributes for the key, to achieve key lifetime and idemptent.

nongfushanquan: if using tidb instead of etcd,how to implement this interface

maxshuang: IMHO, we can design a table with implicit column like revision(bigint unsigned) , startTs(timestamp), ttl(int) to support these attributes, so that `where revision=xxx` can help us achieve idempotent easily.

pkg/metaclient/common.go
```diff
@@ -0,0 +1,147 @@
+package metaclient
+
+// ResponseHeader is common response header
+type ResponseHeader struct {
+	// ClusterId is the ID of the cluster which sent the response.
+	// Framework will generate uuid for every newcoming metastore
+	ClusterID string
```
amyangfei, line 7: How will `ClusterID` be used in client side?

pkg/metaclient/kv.go
amyangfei: Another way is to separate the `put` and `txn`, `txn` does the complex operation, such as setting a key with specific revision, for several reasons
- In many cases idempotent is not necessary for `put` operation. `put` with options makes the `put` operation has too complex semantics
- In DM and TiCDC, they use `etcd.Txn` for correctness serious scenario.

pkg/metaclient/common.go
maxshuang: Just a tag for different metastore backend

PR #140, opened 2022-02-08
runtime, lib: abstract out the interface Runnable

liuzix:
- Abstract out the interface `Runnable` which describes an entity that can be run by the runtime.
- Minor modifications to `Worker` interface.
- Minor adjustment of the type of `WorkerID`.
- Abstract out the interface `Workloader`, which describes an object that can report its workload.


hanfei1991:
Right now the `Runtime` holds `Runnable`. How can we collect the workloads from these Runnable if we don't have `Workload` interface any more?

liuzix:
> Right now the `Runtime` holds `Runnable`. How can we collect the workloads from these Runnable if we don't have `Workload` interface any more?

We have a new interface `Workloader`.

nongfushanquan (approved):

PR #145, opened 2022-02-09
runtime: refine the implementation

liuzix:
Fixed the following problem:
- Adding tasks to channels could block the entire runtime.
- `Workload()` could be called on uninitialized tasks.
- Goroutines are not waited for.
- Runtime cannot be canceled from outside.

Added feature:
- Added error type `ErrRuntimeReachedCapacity`

Other:
- Minor adjustment of code styling.

### DIFF
--- a/executor/worker/runtime.go
+++ b/executor/worker/runtime.go
@@ -6,49 +6,51 @@ import (
 	"time"
 
 	"github.com/edwingeng/deque"
-	"github.com/hanfei1991/microcosm/model"
 	"github.com/pingcap/tiflow/dm/pkg/log"
+	"go.uber.org/atomic"
 	"go.uber.org/zap"
+
+	"github.com/hanfei1991/microcosm/executor/worker/internal"
+	"github.com/hanfei1991/microcosm/model"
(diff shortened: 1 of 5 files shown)


executor/worker/runtime.go
```diff
@@ -65,82 +67,187 @@ func (s *Scheduler) runImpl() {
...
+
+		if !ok {
+			return
+		}
+		// TODO error handling
+		_ = task.Close(ctx)
+		r.taskList.Delete(task.ID())
+
+		r.taskNumMu.Lock()
+		newTaskNum := r.taskNum.Sub(1)
```
hanfei1991, line 147: Isn't atomic sub thread-safe?

liuzix: A lock is used to simplify the implementation of the algorithm below
```
if taskNum < r.capacity {  // L1
    taskNum += 1           // L2
} else {
    // returns error
}
```
Not using a lock will risk other thread's instructions go between L1 and L2.

hanfei1991: That makes sense.

hanfei1991 (approved):

PR #146, opened 2022-02-09
job-master: remove heartbeat in base master, use job master instead

amyangfei:
- Remove `masterClient` and heartbeat logic in `BaseMaster`
- Now job master should implement the interface of `lib.BaseJobMaster`


hanfei1991:
Then is the "masterID" in `lib.Master` still useful?

amyangfei:
> Then is the "masterID" in `lib.Master` still useful?

@hanfei1991 good catch, removed `masterID` from `lib.Master`.

hanfei1991 (approved):

PR #149, opened 2022-02-10
Add Kv client interface

maxshuang:
Add kv client abstract interface and usage examples

### DIFF
--- a/pkg/metaclient/common.go
+++ b/pkg/metaclient/common.go
@@ -1,0 +1,161 @@
+package metaclient
+
+import "fmt"
+
+// ResponseHeader is common response header
+type ResponseHeader struct {
+	// ClusterId is the ID of the cluster which sent the response.
+	// Framework will generate uuid for every newcoming metastore
+	ClusterID string
+}
+
+func (h *ResponseHeader) String() string {
+	return fmt.Sprintf("clusterID:%s;", h.ClusterID)
+}
+
+// Put Response
+type PutResponse struct {
+	Header *ResponseHeader
+}
+
+// Get Response
+type GetResponse struct {
+	Header *ResponseHeader
+	// kvs is the list of key-value pairs matched by the range request.
+	// kvs is empty when count is requested.
+	Kvs []*KeyValue
+}
+
+func (resp *GetResponse) String() string {
+	s := fmt.Sprintf("header:[%s];kvs:[", resp.Header)
+	for _, kv := range resp.Kvs {
+		s += kv.String()
+	}
+
+	s += "];"
+	return s
+}
+
+// Delete Response
+type DeleteResponse struct {
+	Header *ResponseHeader
+}
+
+// Txn Response
+type TxnResponse struct {
+	Header *ResponseHeader
+	// Responses is a list of responses corresponding to the results from applying
+	// success if succeeded is true or failure if succeeded is false.
+	Responses []ResponseOp
+}
+
+type ResponseOp struct {
+	// response is a union of response types returned by a transaction.
+	//
+	// Types that are valid to be assigned to Response:
+	//	*ResponseOp_ResponseRange
+	//	*ResponseOp_ResponsePut
+	//	*ResponseOp_ResponseDeleteRange
+	//	*ResponseOp_ResponseTxn
+	Response isResponseOpResponse
+}
+
+// Using interface to make union
+type isResponseOpResponse interface {
+	isResponseOp()
+}
+
+type ResponseOpResponseGet struct {
+	ResponseGet *GetResponse
+}
+
+type ResponseOpResponsePut struct {
+	ResponsePut *PutResponse
+}
+
+type ResponseOpResponseDelete struct {
+	ResponseDelete *DeleteResponse
+}
+
+type ResponseOpResponseTxn struct {
+	ResponseTxn *TxnResponse
+}
+
+func (*ResponseOpResponseGet) isResponseOp()    {}
+func (*ResponseOpResponsePut) isResponseOp()    {}
+func (*ResponseOpResponseDelete) isResponseOp() {}
+func (*ResponseOpResponseTxn) isResponseOp()    {}
+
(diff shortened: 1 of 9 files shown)


pkg/metaclient/op.go
```diff
@@ -0,0 +1,310 @@
+// Copyright 2016 The etcd Authors
```
liuzix, line 1: If this file is modified based on a third-party work, could you add a link to the original? Also please read the Redistribution section in https://www.apache.org/licenses/LICENSE-2.0

pkg/metaclient/common.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (h *ResponseHeader) String() string {
+	return fmt.Sprintf("clusterID:%s;", h.ClusterID)
+}
+
+// Put Response
+type PutResponse struct {
+	Header *ResponseHeader
+}
```
amyangfei, line 19: If `revision` is used, `PutResponse` should contain `revision` too. Such as the usage
1. put a key for the first time, return the `revision` such as `1000` of this key
2. put the key for another time, with `revision = 1000` to ensure there is concurrent update of this key between step-1 and step-2

maxshuang: Seems that it's easy to get the revision for putResponse when backend is etcd.
But when the backend is TiDB/MySQL, we still need another get to fetch the latest revision.

pkg/metaclient/example.go
```diff
@@ -0,0 +1,174 @@
...
+	//
+	// Txn, forbit nested txn
+	//
+	// current data:
+	//		apple  red
+	//		apple2  green
+	//		ticdc  kv
+	//		dm	   DDL
+	txn := cli.Txn(ctx)
+	txnRsp, err := txn.Do(getOp).Do(putOp).Do(delOp).Commit()
```
amyangfei, line 161: Current `Txn` doesn't support the following scenarios
- etcd like compare, such as `Txn(...).If(...).Then(...).Commit()
- SQL like where condition in op

amyangfei: Given the requirement from DM, https://github.com/pingcap/tiflow/blob/85116e4e6c973c8f82164908bc1041ec56d2fb1a/dm/pkg/shardddl/optimism/operation.go#L121-L123

Supposing the backend is a MySQL like database, we can provide a txn interface that allows user to write raw SQL, or supporting where condition in put op.

In this way we can keep the lightweight of kv interface, which provides simple API such [TiKV RawKV API](https://tikv.org/docs/5.1/develop/rawkv/introduction/), and a customized txn API.

pkg/metaclient/op.go
maxshuang: I have already added the apache license 2.0 file to the project and the license head for op.go file.  PTAL
Should I add a link to the original? I don't find any instruction on it.

liuzix: I think a link to the original may help the maintenance of the code.

maxshuang: Done. PTAL

pkg/metaclient/common.go
```diff
@@ -0,0 +1,161 @@
...
+// ResponseHeader is common response header
+type ResponseHeader struct {
+	// ClusterId is the ID of the cluster which sent the response.
+	// Framework will generate uuid for every newcoming metastore
+	ClusterID string
+}
+
+func (h *ResponseHeader) String() string {
+	return fmt.Sprintf("clusterID:%s;", h.ClusterID)
+}
```
amyangfei, line 14: How often will the `String` function be called, including the `String` method of `GetResponse`
If they are called frequently, use string buffer to concat string instead of using `fmt.Sprintf`

pkg/metaclient/common.go
```diff
@@ -0,0 +1,161 @@
...
+// Put Response
+type PutResponse struct {
+	Header *ResponseHeader
+}
+
+// Get Response
+type GetResponse struct {
+	Header *ResponseHeader
+	// kvs is the list of key-value pairs matched by the range request.
+	// kvs is empty when count is requested.
```
amyangfei, line 25: What does `when count is requested` mean?

pkg/metaclient/kv.go
```diff
@@ -0,0 +1,47 @@
+package metaclient
+
+import "context"
+
+// Txn don't support nested txn
```
amyangfei, line 5:
```suggestion
// Txn doesn't support nested txn
```

pkg/metaclient/kv.go
```diff
@@ -0,0 +1,47 @@
...
+	// WARN: WithRange(end), WithFromKey(), WithPrefix() can't be used at the same time
+	Delete(ctx context.Context, key string, opts ...OpOption) (*DeleteResponse, error)
+
+	// Do applies a single Op on KV without a transaction.
+	// Do is useful when creating arbitrary operations to be issued at a
+	// later time and making intermediate layer for kv; the user can range over the operations, calling Do to
+	// execute them. Get/Put/Delete, on the other hand, are best suited
+	// for when the operation should be issued at the time of declaration.
+	// Same op limit with Put/Get/Delete interface
+	Do(ctx context.Context, op Op) (OpResponse, error)
```
amyangfei, line 43: Is `op` be one of `Put`, `Get` or `Delete`?
Is there anything that we can't do without this `Do` function? Or will we support other extension in `Op`?

pkg/metaclient/common.go
maxshuang: Only for debug log

amyangfei: If using for debug, please don't expose the function, or adding some comments.

pkg/metaclient/kv.go
maxshuang: No.  User will use OpGet/OpPut/OpDelete + WithXXX to generate the corresponding op.
Thanks for reminding. I will  limit the flexibility of Op to limit its usage.  we can support other extension for Op by adding more inner attributes in the future.

amyangfei (approved):

PR #150, opened 2022-02-10
job-manager: support job master auto recover

amyangfei:
- Add a simple job state fsm in job manager.
- Enable job manager `Poll` in `runLeaderLoop` ticker (like a worker runtime to drive `Poll`)
- Support job master failover when job manager is alive.
- `JobMaster` will always use the same `job master id`


servermaster/jobmanager.go
```diff
@@ -126,7 +129,11 @@ func (jm *JobManagerImplV2) InitImpl(ctx context.Context) error {
 
 // Tick implements lib.MasterImpl.Tick
 func (jm *JobManagerImplV2) Tick(ctx context.Context) error {
-	return nil
+	return jm.jobFsm.IterPendingJobs(
```
nongfushanquan, line 132: tick will not be invoked,  tick is designed to be called by poll(in the  worker struct ), and the poll is called by the executor, not by server master

servermaster/jobmanager.go
```diff
@@ -78,8 +77,12 @@ func (jm *JobManagerImplV2) SubmitJob(ctx context.Context, req *pb.SubmitJobRequ
 		resp.Err = errors.ToPBError(err)
 		return resp
 	}
-	resp.JobIdStr = id
 
+	// TODO: data persistence for masterConfig
```
nongfushanquan, line 81: persistence the jobmaster information first, then create worker here ?

servermaster/jobmanager.go
amyangfei: `Poll` is called manually at https://github.com/hanfei1991/microcosm/blob/902e1bcd103f6b6af35c6a4905f0aa3d78e2fa55/servermaster/server.go#L508-L511

servermaster/jobmanager.go
```diff
@@ -140,21 +147,19 @@ func (jm *JobManagerImplV2) OnWorkerDispatched(worker lib.WorkerHandle, result e
...
 func (jm *JobManagerImplV2) OnWorkerOnline(worker lib.WorkerHandle) error {
 	log.L().Info("on worker online", zap.Any("id", worker.ID()))
-	return nil
+	return jm.jobFsm.JobOnline(worker)
 }
 
 // OnWorkerOffline implements lib.MasterImpl.OnWorkerOffline
 func (jm *JobManagerImplV2) OnWorkerOffline(worker lib.WorkerHandle, reason error) error {
 	log.L().Info("on worker offline", zap.Any("id", worker.ID()), zap.Any("reason", reason))
+	jm.jobFsm.JobOffline(worker)
```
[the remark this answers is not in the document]
amyangfei: Does `restart the same jobmaster` mean
- Use the same the job master id
- start job master at the same executor

And what's the difference between creating a new job master and using the old one, is it related to the failover or workers belong to the job master?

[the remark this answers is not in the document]
amyangfei: Imo we should support `CreateWorker` with a given ID, which means the new created job master shares the same ID with old job master, and existing workers can find the new created job master via this ID. I will add some basic failover rules today.

servermaster/jobmanager.go
nongfushanquan: if create worker success, then crash , the job  will run twice

servermaster/jobmanager.go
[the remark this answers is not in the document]
nongfushanquan: yes , the old jobmaster with the same id  will recover and get the persistence information from the meta store.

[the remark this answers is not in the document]
nongfushanquan: the same id , not the same executor.

servermaster/job_fsm.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"go.uber.org/zap"
+)
+
+type jobHolder struct {
+	lib.WorkerHandle
+	*model.JobMasterV2
+}
+
+type JobFsm struct {
```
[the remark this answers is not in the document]
amyangfei: Sure, added comments, the fsm will be refined later.

servermaster/jobmanager.go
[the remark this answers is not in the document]
amyangfei: Updated, job master will use pre-allocated id and this id is persisted in meta store.

servermaster/jobmanager.go
```diff
@@ -69,17 +70,21 @@ func (jm *JobManagerImplV2) SubmitJob(ctx context.Context, req *pb.SubmitJobRequ
 		resp.Err = errors.ToPBError(err)
 		return resp
 	}
+
+	// TODO: data persistence for masterConfig
+
 	// CreateWorker here is to create job master actually
 	// TODO: use correct worker type and worker cost
 	id, err := jm.BaseMaster.CreateWorker(
```
liuzix, line 78: `CreateWorker` can fail at last, because the rpc is done in a background goroutine. Is this situation well handled here?

amyangfei: added a state transition when meeting error in `OnWorkerDispatched`, from `WaitAck` to `Pending`.

lib/master.go
```diff
@@ -408,6 +434,22 @@ func (m *DefaultBaseMaster) registerHandlerForWorker(ctx context.Context, worker
 	return nil
 }
 
+// generateWorkerID assigns a new worker id.
+// When creating a job master, job manager provides a pre allocated ID, in other
+// cases, we need to generate a random WorkerID.
+func (m *DefaultBaseMaster) generateWorkerID(workerType WorkerType, config WorkerConfig) string {
```
nongfushanquan, line 440: As to the. jobmaster , job name  is more reasonable .

amyangfei: It is used in two scenarios,
- job manager generates job master id
- job master generates worker id

Using worker id is more generic.

amyangfei: ptal @nongfushanquan

amyangfei: As for using `job name` as worker id of job master, added a TODO and refine it later. Since the `job name` should be checked unique.

lib/job.go
```diff
@@ -0,0 +1,9 @@
+package lib
+
+// JobMasterV2 holds the config of a job and used for failover
+type JobMasterV2 struct {
+	ID         string      `json:"id"`
+	Tp         WorkerType  `json:"type"`
+	Ext        interface{} `json:"ext"`
```
hanfei1991, line 7: is Ext possible to be a non-string field?

servermaster/job_fsm.go
```diff
@@ -0,0 +1,153 @@
...
+//     |  (with error)             |                    |
+//     |----------------------------------------------->|
+//     |                           |                    |
+//     |                           |                    |
+//     |                           |                    |
+type JobFsm struct {
+	JobStats
+
+	jobsMu      sync.RWMutex
+	pendingJobs map[string]*lib.JobMasterV2
```
hanfei1991, line 51: Why not use lib.workerID as key type?

lib/job.go
amyangfei: will refine in another pr

amyangfei: we should not update too much in one PR.

servermaster/job_fsm.go
amyangfei: Changed to `MasterID`, actually they alias to string actually.

pkg/errors/errors.go
```diff
@@ -60,6 +60,7 @@ var (
 	ErrWorkerOffline      = errors.Normalize("worker is offline: workerID %s", errors.RFCCodeText("DFLOW:ErrWorkerOffline"))
 	ErrWorkerTimedOut     = errors.Normalize("worker heartbeat timed out: workerID %s", errors.RFCCodeText("DFLOW:ErrWorkerTimedOut"))
 	ErrWorkerSuicide      = errors.Normalize("worker has committed suicide due to master having timed out", errors.RFCCodeText("DFLOW:ErrWorkerSuicide"))
+	ErrInvalidJobType     = errors.Normalize("invalid job type: %s", errors.RFCCodeText("DFLOW:ErrInvalidJobType"))
```
hanfei1991, line 63: Could we also print all the valid types in the err message?

amyangfei: will leave a todo in cli

hanfei1991 (approved):
LGTM

PR #151, opened 2022-02-11
fix bugs to run demo

hanfei1991:
After this PR, cvsTask can run as expected. In addition:

- adjust log level
- revert the baseJobMaster to baseMaster

### DIFF
--- a/executor/cvsTask/cvstask.go
+++ b/executor/cvsTask/cvstask.go
@@ -79,6 +79,7 @@ func NewCvsTask(ctx *dcontext.Context, _workerID lib.WorkerID, masterID lib.Mast
 
 func (task *cvsTask) InitImpl(ctx context.Context) error {
 	log.L().Info("init the task  ", zap.Any("task id :", task.ID()))
+	task.status = lib.WorkerStatusNormal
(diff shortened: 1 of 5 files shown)


executor/cvsTask/cvstask.go
```diff
@@ -128,28 +128,28 @@ func (task *cvsTask) CloseImpl(ctx context.Context) error {
...
+			log.L().Error("read data failed", zap.Error(err))
 			if !task.isEOF {
 				task.cancelFn()
 			}
 			return err
 		}
 		if reply.IsEof {
 			log.L().Info("Reach the end of the file ", zap.Any("fileName:", task.srcDir))
-			task.isEOF = true
+			close(task.buffer)
```
nongfushanquan, line 152: if the server crash  in the syncing process, the status should be set error

hanfei1991: I think the status assignment is here https://github.com/hanfei1991/microcosm/pull/151/files#diff-d9aaad6e272ef0ddb3d4b3e96e3d52db823421edb128f729d16bac74fcbc1253R87

hanfei1991:
relate #152

amyangfei (approved):

PR #162, opened 2022-02-15
feat(metaclient(dataflow)): add logical namespace

maxshuang:
close issue: #158


pkg/metaclient/internal/interface.go
```diff
@@ -0,0 +1,15 @@
+package internal
+
+import (
+	"context"
+
+	"github.com/hanfei1991/microcosm/pkg/metaclient"
+)
+
+type KVEx interface {
```
amyangfei, line 9: What does `Ex` mean?

pkg/metaclient/namespace/kv.go
```diff
@@ -0,0 +1,158 @@
...
+
+// NewPrefixKV wraps a KVEx instance so that all requests
+// are prefixed with a given string.
+func NewPrefixKV(kv internal.KVEx, prefix string) metaclient.KV {
+	return &kvPrefix{kv, prefix}
+}
+
+func (kv *kvPrefix) Put(ctx context.Context, key, val string) (*metaclient.PutResponse, error) {
+	if len(key) == 0 {
+		return nil, cerrors.ErrMetaEmptyKey
```
amyangfei, line 40: Should generate the error

pkg/metaclient/namespace/kv.go
```diff
@@ -0,0 +1,158 @@
...
+	r, err := kv.KVEx.Do(ctx, op)
+	if err != nil {
+		return nil, err
+	}
+	put := r.Put()
+	return put, nil
+}
+
+func (kv *kvPrefix) Get(ctx context.Context, key string, opts ...metaclient.OpOption) (*metaclient.GetResponse, error) {
+	if len(key) == 0 && !(metaclient.IsOptsWithFromKey(opts) || metaclient.IsOptsWithPrefix(opts)) {
```
amyangfei, line 52: Why adding `!(metaclient.IsOptsWithFromKey(opts) || metaclient.IsOptsWithPrefix(opts)) ` this check?

pkg/metaclient/namespace/kv.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, cerrors.ErrMetaEmptyKey
+	}
+	r, err := kv.KVEx.Do(ctx, kv.prefixOp(metaclient.OpGet(key, opts...)))
+	if err != nil {
+		return nil, err
+	}
+	get := r.Get()
+	kv.unprefixGetResponse(get)
```
amyangfei, line 60: Supposing the namespace has prefix `/prefix1`, when user queries with `WithFromKey` option, and no key starting from `/prefix1` is matched, and returns a key with prefix `/prefix2`, then user accesses invalid data. 

such as querying `/zz` with `WithFromKey` option. It will return `/prefix2/aa`

```
/prefix1/aa
/prefix1/bb
/prefix2/aa
```

pkg/metaclient/internal/interface.go
maxshuang: Extend the KV interface with Do method to implement the intermidate layer easier

pkg/metaclient/namespace/kv.go
maxshuang: It will not happen.
Actually,  as for your case, internal key span will be [/prefix1, /prefix2)

amyangfei: Which code generates the `/prefix2` part in `[/prefix1, /prefix2)`

maxshuang: prefixOp()--->prefixInterval()

pkg/metaclient/kvclient/client.go
```diff
@@ -0,0 +1,30 @@
...
+// etcdKVClient is the implement of kv interface based on etcd
+// Support namespace isolation and all kv ability
+// etcdImpl -> kvPrefix+etcdClientCloser -> etcdKVClient
+type etcdKVClient struct {
+	metaclient.Closer
+	metaclient.KV
+	leaseID string
+}
+
+func NewEtcdKVClient(config *metaclient.Config, leaseID string) (metaclient.KVClient, error) {
```
lance6716, line 18: the name is leaseID, it makes me think that after the lease is expired, the namespace will be reset 🤔

pkg/metaclient/namespace/kv.go
```diff
@@ -0,0 +1,162 @@
...
+	r, err := kv.KVEx.Do(ctx, op)
+	if err != nil {
+		return nil, err
+	}
+	put := r.Put()
+	return put, nil
+}
+
+func (kv *kvPrefix) Get(ctx context.Context, key string, opts ...metaclient.OpOption) (*metaclient.GetResponse, error) {
+	// Forbit empty key to protect the namespace prefix key
```
lance6716, line 52:
```suggestion
	// Forbid empty key to protect the namespace prefix key
```

and rest of it

pkg/metaclient/namespace/util.go
```diff
@@ -0,0 +1,48 @@
...
+
+func MakeNamespacePrefix(leaseID string) string {
+	return leaseID + "/"
+}
+
+func prefixInterval(pfx string, key, end []byte) (pfxKey []byte, pfxEnd []byte) {
+	pfxKey = make([]byte, len(pfx)+len(key))
+	copy(pfxKey[copy(pfxKey, pfx):], key)
+
+	if len(end) == 1 && end[0] == 0 {
```
lance6716, line 27: not sure the convention of `\0` as the end of key space also applies to other KV store.

lance6716 (commented):
rest lgtm

pkg/metaclient/kvclient/client.go
maxshuang: I will change to tenantID.

pkg/metaclient/namespace/util.go
maxshuang: It's the convention of Op struct. As for the mplement, it need to make some convertion.
E.g., for MySQL, it will be "Where key >= $StartKey".

lance6716 (approved):

PR #163, opened 2022-02-15
master: support passing by master meta ext

amyangfei:
To support the job master recover in job manager failover scenario, we should persist enough information of job master.

This PR refactors the storage of `MasterMetaKVData`, renames `JobMasterV2` to `MasterMetaExt` and uses it as a defined type in `MasterMetaKVData`, the `config` defined in `MasterMetaExt` can be parsed by registered `DeserializeConfig` function.

We can use `MasterMetaExt` to restore a complete job.

### DIFF
--- a/lib/master.go
+++ b/lib/master.go
@@ -110,6 +110,7 @@ type DefaultBaseMaster struct {
 	advertiseAddr string
 	nodeID        p2p.NodeID
 	timeoutConfig TimeoutConfig
+	masterMetaExt *MasterMetaExt
 
 	// components for easier unit testing
 	uuidGen uuid.Generator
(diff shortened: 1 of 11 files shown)


lib/master.go
```diff
@@ -131,10 +132,16 @@ func NewBaseMaster(
 	var (
 		nodeID        p2p.NodeID
 		advertiseAddr string
+		masterMetaExt = &MasterMetaExt{}
 	)
 	if ctx != nil {
```
hanfei1991, line 137: just out of curiosity, when is it nil? I see it's always assigned when "create worker"

amyangfei: maybe in mock module, such as https://github.com/hanfei1991/microcosm/blob/ee6a157c0c101730e6799c2973a863b861034a1b/lib/mock_master_impl.go#L60-L61

hanfei1991 (approved):

PR #170, opened 2022-02-17
persist WorkerStatus in metastore (Part II)

liuzix:
- The `BaseWorker` now persists `WorkerStatus` in metastore, to solve the problem of lost states if both the master and a worker crash at the same time.
- Added a method to `BaseWorker`: 
```go
UpdateStatus(ctx context.Context, status WorkerStatus) error
```

Remains to be done:
- [ ] Fetch all persisted states related to the given master when `BaseMaster` starts up.
- [ ] Test the compatibility with recreating a worker with the same `WorkerID`.


lib/worker.go
```diff
@@ -106,6 +117,21 @@ func (w *DefaultBaseWorker) Workload() model.RescUnit {
 }
 
 func (w *DefaultBaseWorker) Init(ctx context.Context) error {
+	// TODO refine this part
+	poolCtx, cancelPool := context.WithCancel(context.TODO())
```
amyangfei, line 121: Can we use `context.WithCancel(ctx)` here?

lib/worker.go
```diff
@@ -54,6 +56,7 @@ type BaseWorker interface {
 	Close(ctx context.Context) error
 	ID() runtime.RunnableID
 	MetaKVClient() metadata.MetaKV
+	UpdateStatus(ctx context.Context, status WorkerStatus) error
```
amyangfei, line 59: Is this `UpdateStatus` driven by system logic or user logic?

lib/worker.go
liuzix: Yes we can. But we need to clearly specify what context the runtime needs to call `Init` with. The context used to call `Init` cannot be a context with a timeout, or otherwise the background goroutines can be canceled prematurely.

amyangfei (approved):

amyangfei:
A test case failed
```go
--- FAIL: TestHeartBeatPingPongAfterFailover (0.00s)
    worker_manager_test.go:120: 
        	Error Trace:	worker_manager_test.go:120
        	Error:      	"[]" should have 1 item(s), but has 0
        	Test:       	TestHeartBeatPingPongAfterFailover
--- FAIL: TestHeartBeatPingPongAfterCreateWorker (0.00s)
    worker_manager_test.go:63: 
        	Error Trace:	worker_manager_test.go:63
        	Error:      	"[]" should have 1 item(s), but has 0
        	Test:       	TestHeartBeatPingPongAfterCreateWorker
--- FAIL: TestMultiplePendingHeartbeats (0.02s)
    worker_manager_test.go:196: 
        	Error Trace:	worker_manager_test.go:196
        	Error:      	"[]" should have 2 item(s), but has 0
        	Test:       	TestMultiplePendingHeartbeats
--- FAIL: TestWorkerTimedOut (1.02s)
    worker_manager_test.go:439: 
        	Error Trace:	worker_manager_test.go:439
        	Error:      	Condition never satisfied
        	Test:       	TestWorkerTimedOut
FAIL
coverage: 71.3% of statements
FAIL	github.com/hanfei1991/microcosm/lib	3.469s
```

PR #171, opened 2022-02-17
support query job

hanfei1991:
Support query job status. But right now job manager cannot get status from job masters.

TODO: add unit-test


cmd/demoserver/kvdatamake.go
```diff
@@ -195,6 +198,32 @@ func (*DataRWServer) ListFiles(ctx context.Context, folder *pb.ListFilesReq) (*p
...
+	}
+}
+
+func (s *DataRWServer) CheckDir(ctx context.Context, req *pb.CheckDirRequest) (*pb.CheckDirResponse, error) {
+	dir := req.Dir
+	for path := range s.fileWriterMap {
+		name := filepath.Base(path)
+		destPath := filepath.Join(dir, name)
+		cmd := exec.Command("diff", path, destPath)
+		stdout, err := cmd.Output()
```
amyangfei, line 216: Better to use a golang library, such as `https://github.com/sergi/go-diff`, in case of `diff` is not available (no binary or access path is restricted)

hanfei1991:
addressed, PTAL @amyangfei

amyangfei (approved):

PR #176, opened 2022-02-21
lib: fix unstable tests

liuzix:
Fixed two unstable tests:
- TestHeartBeatPingPongAfterFailover
- TestMultiplePendingHeartbeats

### DIFF
--- a/lib/worker_manager_test.go
+++ b/lib/worker_manager_test.go
@@ -201,10 +201,10 @@ func TestMultiplePendingHeartbeats(t *testing.T) {
 
 	replyTime := time.Now()
 	manager.clock.(*clock.Mock).Set(replyTime)
-	offlined, onlined = manager.Tick(ctx, msgSender)
 
 	var totalOnlined int
 	require.Eventually(t, func() bool {
+		offlined, onlined = manager.Tick(ctx, msgSender)
 		totalOnlined += len(onlined)
 		require.Empty(t, offlined)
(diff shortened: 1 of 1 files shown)


lib/worker_manager_test.go
```diff
@@ -199,8 +202,13 @@ func TestMultiplePendingHeartbeats(t *testing.T) {
 	replyTime := time.Now()
 	manager.clock.(*clock.Mock).Set(replyTime)
 	offlined, onlined = manager.Tick(ctx, msgSender)
```
lance6716, line 204: Tick should be included in `Eventually`?

lib/worker_manager_test.go
```diff
@@ -198,9 +201,14 @@ func TestMultiplePendingHeartbeats(t *testing.T) {
 
 	replyTime := time.Now()
 	manager.clock.(*clock.Mock).Set(replyTime)
-	offlined, onlined = manager.Tick(ctx, msgSender)
-	require.Empty(t, offlined)
-	require.Len(t, onlined, 2)
+
+	var totalOnlined int
+	require.Eventually(t, func() bool {
+		offlined, onlined = manager.Tick(ctx, msgSender)
```
lance6716, line 207: `onlined` means justOnlined, very misleading, please add comment for manager.Tick later

lance6716 (approved):

hanfei1991 (approved):

PR #177, opened 2022-02-22
Add  kv client mock

maxshuang:
add mock kv client for KVClient using gomock


pkg/metaclient/example
```diff
@@ -0,0 +1 @@
+kvclient/example.go
```
amyangfei, line 1: What's this file used for?

maxshuang: using mock kv client to show the usage of client interface

amyangfei (approved):

PR #179, opened 2022-02-22
lib/master, lib/worker, lib/base_jobmaster: Refactor master/worker and refine BaseJobMaster

liuzix:
- Renamed `Tick` to `Poll` in `MasterImpl` and `WorkerImpl`.
- Runtime now calls the `Poll` method on the Impls directly, which are in turn responsible for calling the `Poll` on the Bases.
- A similar changed is done to `Init` too, to prevent calling these methods twice in `BaseJobMaster`.
- Miscellaneous changes to accommodate the interface changed.


servermaster/jobmanager.go
```diff
@@ -166,6 +184,7 @@ func (jm *JobManagerImplV2) Tick(ctx context.Context) error {
 
 // OnMasterRecovered implements lib.MasterImpl.OnMasterRecovered
 func (jm *JobManagerImplV2) OnMasterRecovered(ctx context.Context) error {
+	// Deprecated
```
amyangfei, line 187: Can we remove this function from interface, will the failover (such as a master exits and recovers) will be executed in the `Init` function?

jobmaster/cvsJob/cvsJobMaster.go
```diff
@@ -76,7 +78,16 @@ func NewCVSJobMaster(ctx *dcontext.Context, workerID lib.WorkerID, _ lib.MasterI
 	return jm
 }
 
-func (jm *JobMaster) InitImpl(ctx context.Context) error {
+func (jm *JobMaster) Init(ctx context.Context) error {
+	isFirstStartUp, err := jm.BaseMaster.Init(ctx)
```
amyangfei, line 82: Should `CVSJobMaster` holds an `BaseJobMaster` instead of `BaseMaster` after this PR?

PR #186, opened 2022-02-23
lib/*: pass ExtBytes in worker status directly, remove GetExtTypeInfo api

amyangfei:
Change `WorkerStatus` to the following definition, the framework won't parse the `ExtBytes` field into a Go object, only pass through it. And the user logic code can use the byte slice in any way.
```go
type WorkerStatus struct {                                                       
    Code         WorkerStatusCode `json:"code"`                                  
    ErrorMessage string           `json:"error-message"`                         
                                                                                 
    // ExtBytes carries the serialized form of the Ext field.                    
    // See below for more information.                                           
    // Business should parse the raw bytes and decodes into business Go object   
    ExtBytes []byte `json:"ext-bytes"`                                           
} 
```

- In CVSJobmaster, replace `lib.BaseMaster` with `lib.BaseJobMaster`.
- UpdateStatus with a rate limit in cvs task.
- Remove `GetWorkerStatusExtTypeInfo` and `GetJobMasterStatusExtTypeInfo` from all interfaces.
- Fix `lastUnsentStatus` is not reset after it was sent.

### DIFF
--- a/executor/config.go
+++ b/executor/config.go
@@ -36,6 +36,8 @@ var (
 	defaultRPCTimeout        = "3s"
 	defaultDiscoverTicker    = 3 * time.Second
 	defaultMetricInterval    = 15 * time.Second
+
+	defaultCapability int64 = 100 // TODO: make this configurable
 )
 
 // NewConfig creates a new base config for worker.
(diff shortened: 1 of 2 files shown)


amyangfei:
Met a data race, I will fix it later.

nongfushanquan (approved):

Issue #190, opened 2022-02-24
remaining tasks of job manager/master failover


amyangfei:
- [x] job manager persists job master meta before `CreateWorker`, when both job manager and job master crash before job master meta is persisted in job master, the job information will be lost. #193
- [x] Support job pause or finish of job management in job manager
- [x] Support rescheduling pending jobs if no ack is received for long time. #199 
- [x] Refine data structure of `MasterMetaKVData `, #193 https://github.com/hanfei1991/microcosm/blob/d0a23069e069e9ffa1d735597e70e0a7adc3701b/lib/common.go#L121
- [x] Fix the bug that worker treats transferred master as timeout before any heartbeat is received. #191
- [x] Fix the bug that status update handler is not removed after worker is offline. #195
- [x] Fix on master recovered is not called in job manager startup procedure. #196 
- [x] Solve worker message handler ignores late messages because of server master always uses the same id. #198
- [x] Test only job master exits and restarts, the system can failover normally. #355

PR #191, opened 2022-02-25
lib/worker: reset last ack time after master is transfered

amyangfei:
Fix a bug that worker meets master timeout error unexpected. Task-5 in #190

Since worker has a background routine to check master timeout, it maintains the `lastMasterAckedPingTime` internally. Supposing master is transferred, the timeline is as follows

1. worker detects time difference between now and `lastMasterAckedPingTime` is (workerHeartbeatInterval*2, workerTimeoutDuration) (6s, 15s), refresh the master info
2. worker detects time difference between now and `lastMasterAckedPingTime` is larger than `workerTimeoutDuration` (15s) **before any heartbeat response is received from new master**, it treats the master as timeout.


lib/worker.go
```diff
@@ -495,7 +501,7 @@ func (m *masterClient) CheckMasterTimeout(ctx context.Context, clock clock.Clock
 	if sinceLastAcked > 2*m.timeoutConfig.workerHeartbeatInterval &&
 		sinceLastAcked < m.timeoutConfig.workerTimeoutDuration {
```
nongfushanquan, line 505: we should use one configure parameter here instead of two, such as  the wokerTimeDuration  = 2* workerHeartbeatInterval + 3.

amyangfei: Yes, too many parameters make it difficult for debugging, we need to reduce the complexity, but seems only assign `wokerTimeDuration = 2* workerHeartbeatInterval + 3` doesn't reduce the complexity (The check condition still contains these variables).

amyangfei: addressed, ptal @nongfushanquan

liuzix (approved):

nongfushanquan (approved):

PR #193, opened 2022-02-28
master: refine master meta kv storage

amyangfei:
This PR implements task-1 and task-4 in #190

- Move `Config` into `MasterMetaKVData` and use it in both master logic and job manager code.
-  job manager persists job master meta before `CreateWorker`, then in job master init procedure, it can always load a master meta.
- Fix a bug that master registry uses the wrong `masterID` and `workerID` sequence.


lib/common.go
```diff
@@ -124,12 +124,22 @@ type (
 		NodeID      p2p.NodeID `json:"node-id"`
 		Epoch       Epoch      `json:"epoch"`
 		Initialized bool       `json:"initialized"`
```
nongfushanquan, line 126: should the master's status be a member of the meta data?

amyangfei: `Initialized` tells whether the master has ever been initialized, it is used in failover scenario only.

amyangfei: As for the status of master, it has not been determined, leaving a TODO following.

amyangfei: PTAL @nongfushanquan

nongfushanquan (approved):

PR #194, opened 2022-02-28
[e2e_test]: support simple e2e_test

hanfei1991:
after starting docker compose by 
```
docker-compose -f 3m3e.yaml -f demo.yaml up
```
 we can run this simple e2e_test by running `go test` in the e2e directory


cmd/demoserver/kvdatamake.go
```diff
@@ -281,6 +284,7 @@ func (s *DataRWServer) ReadLines(req *pb.ReadLinesRequest, stream pb.DataRWServi
 		case <-s.ctx.Done():
 			return nil
 		default:
+			time.Sleep(time.Millisecond * 5)
```
nongfushanquan, line 287: why sleep here

jobmaster/cvsJob/cvsJobMaster.go
```diff
@@ -114,26 +115,28 @@ func (jm *JobMaster) Tick(ctx context.Context) error {
...
-			log.L().Info("sync file finished ", zap.Any("message", worker.file))
 		} else if status.Code == lib.WorkerStatusError {
 			log.L().Error("sync file failed ", zap.Any("message", worker.file))
 		} else {
 			log.L().Info("worker status abnormal", zap.Any("status", status))
 		}
 	}
 	log.L().Info("cvs job master status  ", zap.Any("id :", jm.workerID), zap.Int64("counter: ", jm.counter))
-	return nil
+	err := jm.BaseJobMaster.UpdateJobStatus(ctx, jm.Status())
```
nongfushanquan, line 134: don't call updatejobstatus frequently, for the updateJobstatus will store the status into the megastore.   Only call it when the status changes

cmd/demoserver/kvdatamake.go
hanfei1991: Slower the read speed to mitigate the pressure of disk.

jobmaster/cvsJob/cvsJobMaster.go
hanfei1991: fixed, PTAL

hanfei1991: updated, PTAL again

nongfushanquan (approved):

PR #197, opened 2022-03-01
lib/worker: workerManager supports loading all worker statuses

liuzix:
- `workerManager` now loads all persisted `WorkerStatus` during initialization. If a heartbeat is not received for a period, some workers can be marked as tombstones.
- `BaseMaster` exposes `IsMasterReady() bool`
- Support removing tombstones by the business logic.


lib/base_jobmaster.go
```diff
@@ -32,6 +32,11 @@ type BaseJobMaster interface {
 	ID() worker.RunnableID
 	UpdateJobStatus(ctx context.Context, status WorkerStatus) error
 
+	// IsMasterReady returns whether the master has received heartbeats for all
+	// workers after a fail-over. If this is the first time the JobMaster started up,
+	// the return value is always true.
+	IsMasterReady() bool
```
nongfushanquan, line 38: why implement this method in jobmaster instead of master

nongfushanquan (approved):

lib/base_jobmaster.go
liuzix: It is implemented in master too. But we need to redefine it in JobMaster too.

PR #200, opened 2022-03-03
lib/master: fix registering handlers after failover

liuzix:
- Fix the master not being able to receive heartbeats from workers after failover.

### DIFF
--- a/lib/worker_manager.go
+++ b/lib/worker_manager.go
@@ -37,6 +37,7 @@ type workerManager interface {
 	GetWorkers() map[WorkerID]WorkerHandle
 	GetStatus(id WorkerID) (*WorkerStatus, bool)
 	CheckStatusUpdate(ctx context.Context) error
+	RegisterHandler(ctx context.Context, workerID WorkerID) error
 }
 
 type workerManagerFsmState = int32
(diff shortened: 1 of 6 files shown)


lib/worker_manager.go
```diff
@@ -311,7 +387,7 @@ func (m *workerManagerImpl) HandleHeartbeat(msg *HeartbeatPingMessage, fromNode
 	log.L().Debug("received heartbeat", zap.Any("msg", msg))
 	workerInfo, ok := m.workerInfos[msg.FromWorkerID]
 	if !ok {
-		if m.initialized {
+		if m.fsmState.Load() != workerManagerWaitingHeartbeats {
```
amyangfei, line 390: If the `fsmState` has been `workerManagerNormal`, and a worker with the same id restarts, will the worker id in `m.workerInfos`?

liuzix: In the case, the restart must have been triggered by `CreateWorker` on this master, so the worker info is added already.

lib/worker_manager.go
```diff
@@ -173,6 +189,17 @@ func (m *workerManagerImpl) asyncLoadAllWorkers(ctx context.Context) error {
 			// If the worker is proven alive, it will be removed from the
 			// tombstone list.
```
nongfushanquan, line 190: a topic for the tombstone worker will be created

lib/worker_manager.go
```diff
@@ -184,6 +211,55 @@ func (m *workerManagerImpl) asyncLoadAllWorkers(ctx context.Context) error {
...
+		return errors.Trace(err)
+	}
+	if !ok {
+		log.L().Panic("duplicate handler",
+			zap.String("topic", topic))
+	}
+	return nil
+}
+
+func (m *workerManagerImpl) unregisterHandler(workerID WorkerID) error {
```
nongfushanquan, line 245: it is an ugly implement about the message topic in the master now

PR #203, opened 2022-03-07
fix a demo race bug on file writer

hanfei1991:
1. flush the file which this thread is in charge of
2. fix stop command

### DIFF
--- a/ansible/roles/executor/templates/stop.sh.j2
+++ b/ansible/roles/executor/templates/stop.sh.j2
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-ps -e | grep executor-{{ cluster_name }} | awk '{ print $1 }' | xargs -I _ kill -9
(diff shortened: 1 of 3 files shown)


ansible/roles/executor/templates/stop.sh.j2
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-ps -e | grep executor-{{ cluster_name }} | awk '{ print $1 }' | xargs -I _ kill -9
+ps -e | grep executor-{{ cluster_name }} | awk '{ print $1 }' | xargs -r kill -9
```
amyangfei, line 3: Can we use `kill` instead of `kill -9` here?

hanfei1991: addressed

amyangfei (approved):

PR #210, opened 2022-03-08
worker: support worker finish in one direction

amyangfei:
This PR supports worker to exit normally when its worker is done. The changes include
- Remove `Initialized` from `MasterMetaKVData` and add `StatusCode`, which combines the initialization semantic and job master working status.
- Add an `Exit` API in both `BaseJobMaster` and `BaseWorker`, when job master or worker (in user logic code) wants to exit and marks job as finished, it can call the `Exit` function directly.
- After `Exit` is called, base worker will update worker status to finished, and job master will mark master meta to finished to.
- When job manager failovers, it will skip all finished job masters.

### DIFF
--- a/lib/base_jobmaster.go
+++ b/lib/base_jobmaster.go
@@ -32,6 +32,12 @@ type BaseJobMaster interface {
 	ID() worker.RunnableID
 	UpdateJobStatus(ctx context.Context, status WorkerStatus) error
 
+	// Exit should be called when job master (in user logic) wants to exit
+	// - If err is nil, it means job master exits normally
+	// - If err is not nil, it means job master meets error, and after it exits
+	//   it will be failover.
+	Exit(ctx context.Context, status WorkerStatus, err error) error
+
 	// IsMasterReady returns whether the master has received heartbeats for all
 	// workers after a fail-over. If this is the first time the JobMaster started up,
 	// the return value is always true.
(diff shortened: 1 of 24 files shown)


lib/base_jobmaster.go
```diff
@@ -203,6 +209,17 @@ func (d *DefaultBaseJobMaster) IsMasterReady() bool {
 	return d.master.IsMasterReady()
 }
 
+func (d *DefaultBaseJobMaster) Exit(ctx context.Context, status WorkerStatus, err error) error {
```
nongfushanquan, line 212: how to make sure this method is called by the jobmaster, if not ,the jobmanager can't recognized which jobmaster should enter the failover process

amyangfei: In current design it must be called from job master (user logic)
- If user logic doesn't call `Exit`, the job *can't finish*
- Besides, if job master meets error and exits, the job manager can failover it as before.

nongfushanquan: it‘s not reasonable now ,since as designed ,the executor will cal the tick method of the worker , if the tick method return error , the executor will assume the worker in the finished status or error status. then store the worker's status and remove it from the runtime

amyangfei: executor doesn't access the kv meta of worker status

amyangfei: The `Exit` is designed to be called in `Tick` of worker / job master

nongfushanquan (approved):
The worker status and jobmaster status should to be refined later

PR #218, opened 2022-03-10
executor/runtime: implement a new runtime with separate goroutines per task

liuzix:
- Created `TaskRunner` which uses dedicated goroutines to run each task.

### DIFF
--- a/executor/worker/task_runner.go
+++ b/executor/worker/task_runner.go
@@ -1,0 +1,204 @@
+package worker
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/hanfei1991/microcosm/model"
+	"github.com/pingcap/errors"
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"go.uber.org/zap"
+	"golang.org/x/sync/semaphore"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/hanfei1991/microcosm/executor/worker/internal"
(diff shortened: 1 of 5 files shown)


executor/worker/task_runner.go
```diff
@@ -0,0 +1,204 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Trace(ctx.Err())
+		case task := <-r.inQueue:
+			if task == nil {
+				return derror.ErrRuntimeIsClosed.GenWithStackByArgs()
+			}
+			if err := r.onNewTask(ctx, task); err != nil {
+				log.L().Warn("Failed to launch task", zap.Error(err))
```
hanfei1991, line 86: We can also print id here

executor/worker/task_runner.go
```diff
@@ -0,0 +1,206 @@
...
+	r.wg.Add(1)
+	go func() {
+		defer r.wg.Done()
+
+		defer func() {
+			err := t.Close(ctx)
+			log.L().Info("Task Closed", zap.String("id", t.ID()), zap.Error(err))
+			t.OnStopped()
+
+			if _, ok := r.tasks.LoadAndDelete(t.ID()); !ok {
```
nongfushanquan, line 192: Add some about the tasks' number

nongfushanquan (commented):
remove the old runtime code

liuzix:
> remove the old runtime code

I will do this in the next PR.

nongfushanquan (approved):

PR #225, opened 2022-03-15
Some prepare work to support cvs job failover

hanfei1991:
This change is mainly in demo-server. Right now we read and write to kv-db in order to
1. read from middle
2. repeated write


executor/cvsTask/cvstask.go
```diff
@@ -31,24 +30,25 @@ type strPair struct {
...
 type Config struct {
+	Idx      int    `json:"Idx"`
 	SrcHost  string `json:"SrcHost"`
-	SrcDir   string `json:"SrcDir"`
 	DstHost  string `json:"DstHost"`
-	DstDir   string `json:"DstDir"`
-	StartLoc int64  `json:"StartLoc"`
+	DstDir   string `json:"DstIdx"`
+	StartLoc string `json:"StartLoc"`
 }
```
amyangfei, line 38: Should we update the demo task config in `sample/config/demo.json`

hanfei1991: Good catch. Let me think about how to launch demo job by master-client ...

hanfei1991: Finished, We can still use master-client to test locally, which need to add flags to demo-server, such as `./bin/demoserver -d /datax -a 0.0.0.0:1234 -r 1000000`  PTAL

cmd/demoserver/kvdata_test.go
```diff
@@ -11,21 +13,30 @@ import (
...
 
-const (
-	demoAddr = "127.0.0.1:1234"
-	dirName  = "/tmp/df_test"
-)
-
 func TestDemoLogic(t *testing.T) {
+	DemoAddress = "127.0.0.1:1234"
+	DemoDir = "/tmp/data"
+	WtDir := "/tmp/data1"
```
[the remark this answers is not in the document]
hanfei1991: This dirs are created by pkg/db/peddle.go in every "openPeddle", no matter it is created or not. 
Is it still nesessary?

amyangfei (approved):

PR #228, opened 2022-03-15
lib/worker: add messasge channel between master and worker

amyangfei:
Support pause a running job

- Add a message channel between master and worker, worker registers `WorkerStatusChangeRequestTopic` for status change request from master
- Add ` OnMasterMessage(topic p2p.Topic, message p2p.MessageValue) error` API to `WorkerImpl`
- Add a `nonblocking` parameter in `SendMessage` function of worker handler
- Add `pause-job` command in ctl
- **User logic should handle the status change request message, in job master, it should update job status and call `Exit` when needed. In worker it can stop necessary tasks when the user logic think it is safe to exit.**


pkg/p2p/message_sender.go
```diff
@@ -28,6 +31,19 @@ func NewMessageSender(router MessageRouter) MessageSender {
...
+// SendToNodeB implements MessageSender.SendToNodeB
+func (m *messageSenderImpl) SendToNodeB(
+	ctx context.Context, targetNodeID NodeID, topic Topic, message interface{},
+) error {
+	client := m.router.GetClient(targetNodeID)
+	if client == nil {
+		return nil
+	}
+
+	_, err := client.SendMessage(ctx, topic, message)
```
liuzix, line 43: The blocking send might have a performance issue. Can you record a TODO?

liuzix:
Even with a blocking send, there is no guarantee that the message is received. For a check for reliable delivery, you need to check the `Ack` status. Can you make sure there is no problem in the use case in this PR?

lib/worker.go
```diff
@@ -406,6 +402,23 @@ func (w *DefaultBaseWorker) initMessageHandlers(ctx context.Context) error {
...
 	}
+
+	topic = WorkerStatusChangeRequestTopic(w.masterID, w.id)
+	ok, err = w.messageHandlerManager.RegisterHandler(
+		ctx,
+		topic,
+		&StatusChangeRequest{},
+		func(sender p2p.NodeID, value p2p.MessageValue) error {
+			msg := value.(*StatusChangeRequest)
+			return w.Impl.OnMasterMessage(topic, msg)
```
nongfushanquan, line 413: if the onMasterMessage costs a long time, the message server won't work

amyangfei:
> Even with a blocking send, there is no guarantee that the message is received. For a check for reliable delivery, you need to check the `Ack` status. Can you make sure there is no problem in the use case in this PR?

Is the `ack` status exposed from p2p library, how to check it?

lib/worker.go
amyangfei: Good catch, I will add an asynchronous mechanism

amyangfei:
PTAL @nongfushanquan

nongfushanquan (approved):

PR #231, opened 2022-03-16
jobmanager: fix job is not removed from ack list if it finishes before online

amyangfei:
- A job master could finish before `OnWorkerOnline` is called, in this scenario the job should be removed from wait ack list when `OnWorkerOffline` is called.
- Refine the state transition diagram of job manager

### DIFF
--- a/servermaster/job_fsm.go
+++ b/servermaster/job_fsm.go
@@ -227,18 +227,19 @@ func (fsm *JobFsm) JobOffline(worker lib.WorkerHandle, needFailover bool) {
 	defer fsm.jobsMu.Unlock()
 
 	job, ok := fsm.onlineJobs[worker.ID()]
-	if !ok {
+	if ok {
+		delete(fsm.onlineJobs, worker.ID())
+	} else {
 		job, ok = fsm.waitAckJobs[worker.ID()]
 		if !ok {
 			log.L().Warn("unknown worker, ignore it", zap.String("id", worker.ID()))
 			return
 		}
+		delete(fsm.waitAckJobs, worker.ID())
 	}
 	if needFailover {
 		fsm.pendingJobs[worker.ID()] = job.MasterMetaKVData
 	}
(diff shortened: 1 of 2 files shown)


servermaster/job_fsm.go
```diff
@@ -22,32 +22,45 @@ type jobHolder struct {
...
+//     |                           |   (finish)         |               |
+//     |                           |----------------------------------->|
+//     |                           |                    |               |
+//     | Master                    |                    |               |
+//     |  .OnWorkerOffline         |                    |               |
+//     |  (failover)               |                    |               |
+//     |----------------------------------------------->|               |
+//     |                           |                    |               |
+//     | Master                    |                    |               |
+//     |  .OnWorkerOffline         |                    |               |
```
maxshuang, line 49: Can the worker check the first heartbeat pong before it do some work?
Seems strange that the worker state is marked finish before it's online.

amyangfei: They are separated logic, the fsm can't control whether heartbeat is received.
Besides the worker could finish before any heartbeat is sent.

nongfushanquan: this problem should be solved by the worker

amyangfei: Worker can try to send one heartbeat during initialization, but we can't ensure the pong message is received, which can infer `OnWorkerOnline` is called in master, this is caused by an underlying assumption: **Worker should continue to work without confirming the heartbeat between master is normal.**
Supposing the following scenario
- During worker initialization, the heartbeat ping message is blocking or heartbeat ping is sent, but heartbeat pong is not received, should the worker continues to work now?

nongfushanquan: I think the worker should continue work

nongfushanquan (approved):
approve this pr ,then we should refine the relationship between the status and heartbeat in the worker

PR #242, opened 2022-03-21
pkg/dataset: implement a generic CRUD helper

liuzix:
- Implemented `pkg/dataset` using generics.


nongfushanquan (approved):

pkg/dataset/dataset_test.go
```diff
@@ -0,0 +1,67 @@
...
+	Value int
+}
+
+func (r *record) GetID() string {
+	return r.RID
+}
+
+func TestDataSetBasics(t *testing.T) {
+	mockKV := mock.NewMetaMock()
+	dataset := NewDataSet[record, *record](mockKV, adapter.MasterInfoKey)
```
lance6716, line 25: Can we use one type parameter? (I'm learning generic spec now)

liuzix: The two parameters are so designed that `GetID()` is forced to have a pointer receiver.

lance6716: https://go.googlesource.com/proposal/+/refs/heads/master/design/43651-type-parameters.md#pointer-method-example

from here I learned we can use

```golang
	mockKV := mock.NewMetaMock()
	dataset := NewDataSet[record](mockKV, adapter.MasterInfoKey)
	err := dataset.Upsert(context.TODO(), &record{
		RID:   "123",
		Value: 123,
	})
	require.NoError(t, err)

	err = dataset.Upsert(context.TODO(), &record{
		RID:   "123",
		Value: 456,
	})
	require.NoError(t, err)
```

But my GoLand can't recognise it

lance6716 (approved):

PR #245, opened 2022-03-22
lib/worker: fix data race when updating worker status synchronously

amyangfei:
- Separate the `SendStatus` into two sub functions, one of them will persist worker status first in a synchronous way, the other function still keeps the asynchronous style.
- Refactor `SafeSendStatus`, it will try to send status in the synchronous way until it is sucessful. This fixes a bug that when there exits a pending message in status sending buffer, the persist worker status could be overlapped with old status.
- Fix in demo job, worker status is not updated to correct value before exiting.

### DIFF
--- a/jobmaster/cvsJob/cvsJobMaster.go
+++ b/jobmaster/cvsJob/cvsJobMaster.go
@@ -151,6 +151,7 @@ func (jm *JobMaster) Tick(ctx context.Context) error {
 		log.L().Info("cvs job master status", zap.Any("id", jm.workerID), zap.Int64("counter", jm.counter), zap.Any("status", jm.getStatusCode()))
 	}
 	if jm.getStatusCode() == lib.WorkerStatusStopped {
+		jm.setStatusCode(lib.WorkerStatusStopped)
 		log.L().Info("cvs job master stopped")
 		return jm.BaseJobMaster.Exit(ctx, jm.Status(), nil)
 	}
--- a/jobmaster/cvsJob/cvsJobMaster.go
+++ b/jobmaster/cvsJob/cvsJobMaster.go
@@ -151,7 +151,6 @@ func (jm *JobMaster) Tick(ctx context.Context) error {
 		log.L().Info("cvs job master status", zap.Any("id", jm.workerID), zap.Int64("counter", jm.counter), zap.Any("status", jm.getStatusCode()))
 	}
 	if jm.getStatusCode() == lib.WorkerStatusStopped {
-		jm.setStatusCode(lib.WorkerStatusStopped)
 		log.L().Info("cvs job master stopped")
 		return jm.BaseJobMaster.Exit(ctx, jm.Status(), nil)
 	}
--- a/lib/status.go
+++ b/lib/status.go
@@ -24,6 +24,11 @@ const (
 	senderFSMSending
 )
 
+const (
+	safeSendStatusTimeout  = 10 * time.Second
+	safeSendStatusInterval = 10 * time.Millisecond
+)
+
 // StatusSender is used for a worker to send its status to its master.
 type StatusSender struct {
 	workerID WorkerID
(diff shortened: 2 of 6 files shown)


lib/worker.go
```diff
@@ -220,16 +220,14 @@ func (w *DefaultBaseWorker) doPreInit(ctx context.Context) error {
 		return errors.Trace(err)
 	}
 
+	if err := w.statusSender.SafeSendStatus(ctx, WorkerStatus{Code: WorkerStatusInit}); err != nil {
```
nongfushanquan, line 223: what does the workerstatusInit mean to the worker?

jobmaster/cvsJob/cvsJobMaster.go
```diff
@@ -151,6 +151,7 @@ func (jm *JobMaster) Tick(ctx context.Context) error {
 		log.L().Info("cvs job master status", zap.Any("id", jm.workerID), zap.Int64("counter", jm.counter), zap.Any("status", jm.getStatusCode()))
 	}
 	if jm.getStatusCode() == lib.WorkerStatusStopped {
+		jm.setStatusCode(lib.WorkerStatusStopped)
```
nongfushanquan, line 154: if jm.getStatuscode() == lib.workersatusStopped , why need setStatusCode again

lib/status.go
```diff
@@ -97,19 +97,45 @@ func (s *StatusSender) setLastUnsentStatus(status *WorkerStatus) {
...
 
-// SendStatus is used by the business logic in a worker to notify its master
+// AsyncSendStatus wraps sendStatusWrapper with asyncPersist=true
+func (s *StatusSender) AsyncSendStatus(ctx context.Context, status WorkerStatus) error {
+	return s.sendStatusWrapper(ctx, status, true /* asyncPersist */)
+}
+
+// SyncSendStatus wraps sendStatusWrapper with asyncPersist=false
+func (s *StatusSender) SyncSendStatus(ctx context.Context, status WorkerStatus) error {
+	return s.sendStatusWrapper(ctx, status, false /* asyncPersist */)
```
nongfushanquan, line 128: should  be  /*syncPersist*/  here ?

lib/status.go
```diff
@@ -97,19 +97,45 @@ func (s *StatusSender) setLastUnsentStatus(status *WorkerStatus) {
...
 func (s *StatusSender) SafeSendStatus(ctx context.Context, status WorkerStatus) error {
-	if err := s.workerMetaClient.Store(ctx, s.workerID, &status); err != nil {
-		return err
+	for {
+		select {
+		case <-ctx.Done():
+			return errors.Trace(ctx.Err())
+		default:
+		}
+		err := s.SyncSendStatus(ctx, status)
```
nongfushanquan, line 109: max retry time should be set here?

nongfushanquan (commented):
the worker just expect to send status to the master, eg , update ext information to the master, and the status keep unchanged. In this case ,it shouldn't store the meta.

lib/status.go
amyangfei: removed, it is the variable name actually.

jobmaster/cvsJob/cvsJobMaster.go
amyangfei: removed

lib/worker.go
amyangfei: Keep the original logic, after InitImpl is called, change worker status to `WorkerStatusInit`. But this needs to be refactored later, since user could have updated status in `InitImpl`, then the worker status will be overwritten in `doPostInit`

lib/status.go
amyangfei: added default timeout

amyangfei:
> the worker just expect to send status to the master, eg , update ext information to the master, and the status keep unchanged. In this case ,it shouldn't store the meta.

The running status field needs to be persisted.

amyangfei:
ptal @nongfushanquan

nongfushanquan (approved):

lib/status.go
nongfushanquan: where is timeout ?

amyangfei: It is mixed in context
https://github.com/hanfei1991/microcosm/blob/af60dd6d7fd4b43cd2137219da168364e8d6c7d2/lib/status.go#L107-L116

PR #247, opened 2022-03-22
fix(deployment): fix metastore config and deployment error

maxshuang:
replace the etcd listening host and fix config error

### DIFF
--- a/sample/config/etcd.yml
+++ b/sample/config/etcd.yml
@@ -58,7 +58,7 @@ discovery-proxy:
 discovery-srv:
 
 # Initial cluster configuration for bootstrapping.
-initial-cluster:  'dataflow-etcd=http://dataflow-etcd:2380"'
+initial-cluster:  'dataflow-etcd=http://localhost:2380'
 
 # Initial cluster token for the etcd cluster during bootstrap.
 initial-cluster-token: 'etcd-cluster'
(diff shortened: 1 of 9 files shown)


sample/config/etcd.yml
```diff
@@ -39,11 +39,11 @@ cors:
 
 # List of this member's peer URLs to advertise to the rest of the cluster.
 # The URLs needed to be a comma-separated list.
-initial-advertise-peer-urls: http://localhost:2380
+initial-advertise-peer-urls: http://0.0.0.0:2380
 
 # List of this member's client URLs to advertise to the public.
 # The URLs needed to be a comma-separated list.
-advertise-client-urls: http://localhost:2379
+advertise-client-urls: http://0.0.0.0:2379
```
hanfei1991, line 46: These advertise urls will be a problem to use 0.0.0.0 or 127.0.0.1 when we deploy 3 etcd nodes. Advertise address is what we expose to outside world, which should be its domain address.

But for now it ok to me, you can leave a to-do comment

servermaster/server.go
```diff
@@ -345,6 +345,7 @@ func (s *Server) RegisterMetaStore(
 func (s *Server) QueryMetaStore(
 	ctx context.Context, req *pb.QueryMetaStoreRequest,
 ) (*pb.QueryMetaStoreResponse, error) {
+	log.L().Info("receive query metastore request", zap.Any("req", req.Tp))
```
amyangfei, line 348:
```suggestion
```

sample/config/etcd.yml
amyangfei: merge this pr first, better to optimize the etcd config in another PR.

PR #252, opened 2022-03-23
pkg/externalresource: ResourceManager (part II)

liuzix:
- Implemented `manager.Service`
- Refined other related components.


pkg/externalresource/manager/service.go
```diff
@@ -54,5 +85,121 @@ func (s *Service) CreateResource(
 		return nil, st.Err()
 	}
 
+	// Note that the cache is updated AFTER the metastore is updated.
+	if _, exists := s.cache[request.GetResourceId()]; exists {
```
lance6716, line 107: As we discussed in group chat, pkg/externalresource/manager.(*Service).CreateResource doesn't have functionality of forwarding to leader, so it can't be directly embeded into servermaster.(*Server). We still need to write two place when add RPC.

pkg/externalresource/manager/service.go
```diff
@@ -54,5 +85,121 @@ func (s *Service) CreateResource(
...
+		getResourceDuration := time.Since(startTime)
+
+		logger.Info("Resource meta fetch completed", zap.Duration("duration", getResourceDuration))
+
+		if err != nil {
+			return "", false, err
+		}
+		if !exists {
+			return "", false, derror.ErrResourceDoesNotExist.GenWithStackByArgs(id)
+		}
```
lance6716, line 162: can update cache

liuzix: I'm still thinking whether it is necessary to do negative caching

pkg/externalresource/manager/service.go
```diff
@@ -2,37 +2,149 @@ package manager
...
+				ErrorCode:  pb.ResourceErrorCode_ResourceManagerInternalError,
+				StackTrace: errors.ErrorStack(err),
+			})
+			if stErr != nil {
+				return nil, stErr
+			}
+			return nil, st.Err()
+		}
+		if !exists {
+			st, stErr := status.New(codes.Internal, "resource manager error").WithDetails(&pb.ResourceError{
```
amyangfei, line 91: Is it better to use `codes.NotFound` here, ditto for L105

pkg/externalresource/manager/service.go
```diff
@@ -2,37 +2,149 @@ package manager
...
+		return nil, status.Error(codes.Unavailable, "ResourceManager is initializing")
+	}
+
 	if !s.mu.Lock(ctx) {
 		return nil, status.Error(codes.Canceled, ctx.Err().Error())
 	}
 	defer s.mu.Unlock()
 
-	ok, err := s.accessor.CreateResource(ctx, &resourcemeta.ResourceMeta{
+	if _, exists := s.cache[request.GetResourceId()]; exists {
```
amyangfei, line 130: Why there is not check for cache miss here, but the `QueryResource` function has

pkg/externalresource/manager/service.go
```diff
@@ -54,5 +167,199 @@ func (s *Service) CreateResource(
...
+		log.L().Panic("Cache is inconsistent",
+			zap.String("resource-id", request.GetResourceId()))
+	}
+	s.cache[request.GetResourceId()] = resourceRecord
+
+	// TODO: handle the case where resourceRecord.Deleted == true
 	return &pb.CreateResourceResponse{}, nil
 }
+
+func (s *Service) GetPlacementConstraint(
```
amyangfei, line 181: Would you please add comments about what this function does?

pkg/externalresource/broker/storage_handle.go
```diff
@@ -73,12 +73,26 @@ func (f *Factory) NewHandleForLocalFile(
...
+	var creatorWorkerID string
+	if exists {
+		creatorWorkerID = record.Worker
+	} else {
+		creatorWorkerID = workerID
+	}
+	filePath := filepath.Join(getWorkerDir(f.config, creatorWorkerID), suffix)
 	log.L().Info("Using local storage with path", zap.String("path", filePath))
 
 	ls, err := brStorage.NewLocalStorage(filePath)
```
lance6716, line 98: this function is exported for test, not sure if we should find a formal one

pkg/externalresource/manager/service.go
```diff
@@ -54,5 +167,199 @@ func (s *Service) CreateResource(
 		return nil, st.Err()
 	}
 
+	// Note that the cache is updated AFTER the metastore is updated.
+	if _, exists := s.cache[request.GetResourceId()]; exists {
+		log.L().Panic("Cache is inconsistent",
```
lance6716, line 172: we use upsert for metastore, so even if we panic here, the ground truth has been changed and after another resource manager takes over the service will continue. It's OK that we only warn here?

pkg/externalresource/manager/service.go
lance6716: (and the meaning of returned bool, it's true or false for no constraint?)

pkg/externalresource/manager/service.go
```diff
@@ -54,5 +167,199 @@ func (s *Service) CreateResource(
...
+	if !exists {
+		// Note that although we are not doing cache eviction,
+		// a miss is still a possibility given that we might have
+		// a successful write to metastore but due to network problem
+		// we have treated that write as an error.
+		logger.Info("Resource cache miss")
+		var err error
+
+		startTime := time.Now()
+		record, exists, err = s.accessor.GetResource(ctx, id)
```
lance6716, line 217: here we may extract a helper function to reduce duplication with QueryResource

pkg/externalresource/manager/service.go
```diff
@@ -54,5 +167,199 @@ func (s *Service) CreateResource(
...
+		zap.String("executor-id", string(executorID)))
+	return nil
+}
+
+func (s *Service) StartBackgroundWorker() {
+	ctx, cancel := context.WithCancel(context.Background())
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+		<-s.cancelCh
```
lance6716, line 262: why don't we let caller directly call cancel? it's concurrent-safe

pkg/externalresource/manager/service.go
liuzix: Because in the case of a cache miss, the potential name conflict situation is directly handled by the `accessor‘.

pkg/externalresource/manager/service.go
liuzix: This line seems to be longer needed.

pkg/externalresource/manager/service.go
liuzix: Yes, good idea.

pkg/externalresource/manager/service.go
liuzix: Storing a channel seems more elegant considering we need to handle race with `Stop()`.

lance6716 (approved):

PR #256, opened 2022-03-24
fix(master): fix epoch update error

### LINKED ISSUE #255: Epoch of master meta is not updated in failover scenario
Given the following scenario

1. Create a master and an executor, create a job
2. Restart executor
3. Existing job will be re-created by job manager as following
   https://github.com/hanfei1991/microcosm/blob/94bc258770e8dc43e2b52ef96e42554df1dcb45b/servermaster/jobmanager.go#L232-L238

Although epoch is updated in step-3 (this behavior is since #232), it only affects the in-memory value of master meta in job manager, the persisted master meta is not updated.

After the new master re-spawns in executor, it loads master meta from metastore and uses the old epoch, which will break the semantic of epoch.

maxshuang:
close  #255

Using etcdCli in executor to generate increasing epoch for master


lib/master.go
```diff
@@ -201,6 +206,7 @@ func NewBaseMaster(
 		// [TODO] use tenantID if support muliti-tenant
 		userMetaKVClient: kvclient.NewPrefixKVClient(params.UserRawKVClient, tenant.DefaultUserTenantID),
 		deps:             ctx.Deps(),
+		epochGen:         epoch.NewEpochGenerator(params.InnerEtcdClient),
```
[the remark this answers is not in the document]
maxshuang: Seems ok to provide such interface.

pkg/meta/kvclient/etcdkv/etcd_client_test.go
```diff
@@ -528,6 +529,48 @@ func (suite *SuiteTestEtcd) TestTxn() {
...
+			epoch, err = kvcli.GenEpoch(ctx)
+			require.Nil(t, err)
+			require.Greater(t, epoch, oldEpoch)
+		}()
+	}
+
+	wg.Wait()
+	lastEpoch, err := kvcli.GenEpoch(ctx)
+	require.Nil(t, err)
+	require.GreaterOrEqual(t, lastEpoch, int64(0))
```
amyangfei, line 569: This check is useless, since it is weaker than check in L571

pkg/meta/kvclient/mock/simple_mockclient_test.go
```diff
@@ -153,3 +157,42 @@ func TestMockBasicKV(t *testing.T) {
...
+			epoch, err = kvcli.GenEpoch(ctx)
+			require.Nil(t, err)
+			require.Greater(t, epoch, oldEpoch)
+		}()
+	}
+
+	wg.Wait()
+	lastEpoch, err := kvcli.GenEpoch(ctx)
+	require.Nil(t, err)
+	require.GreaterOrEqual(t, lastEpoch, int64(0))
```
amyangfei, line 187: ditto

pkg/meta/kvclient/etcdkv/etcd_client_test.go
```diff
@@ -528,6 +529,48 @@ func (suite *SuiteTestEtcd) TestTxn() {
...
+
```
amyangfei, line 570:
```suggestion
```

pkg/meta/kvclient/mock/simple_mockclient_test.go
```diff
@@ -153,3 +157,42 @@ func TestMockBasicKV(t *testing.T) {
...
+
```
amyangfei, line 188:
```suggestion
```

amyangfei (approved):

PR #257, opened 2022-03-25
lib: refactor status handling

liuzix:
- Extracted status handling logic to a separate package.
- Added the status content to the p2p message so that metastore is no longer read from each time the status is updated.


jobmaster/cvsJob/cvsJobMaster.go
```diff
@@ -26,6 +26,8 @@ import (
 	"go.uber.org/atomic"
 )
 
+var _ lib.JobMasterImpl = &JobMaster{}
```
nongfushanquan, line 29: why need this var ?

lib/base_jobmaster.go
```diff
@@ -63,6 +63,7 @@ type JobMasterImpl interface {
 	OnWorkerOnline(worker WorkerHandle) error
 	OnWorkerOffline(worker WorkerHandle, reason error) error
 	OnWorkerMessage(worker WorkerHandle, topic p2p.Topic, message interface{}) error
+	OnWorkerStatusUpdated(worker WorkerHandle, newStatus *WorkerStatus) error
```
nongfushanquan, line 66: is it necessary  to provide this method?

jobmaster/cvsJob/cvsJobMaster.go
liuzix: This is for testing purpose. We can move it somewhere else

lib/base_jobmaster.go
liuzix: The business logic might want to handle changes to the `ext` field. Providing this method seems the most convenient.

lib/statusutil/writer.go
```diff
@@ -0,0 +1,130 @@
...
+			return
+		}
+		log.L().Warn("UpdateStatus failed",
+			zap.String("worker-id", w.workerID),
+			zap.String("master-id", w.masterInfo.MasterID()),
+			zap.String("master-node", w.masterInfo.MasterNode()),
+			zap.Int64("master-epoch", w.masterInfo.Epoch()))
+	}()
+
+	if reflect2.IsNil(w.lastStatus) || newStatus.HasChanged(w.lastStatus) {
```
amyangfei, line 61: Where do we update `w.lastStatus`?

nongfushanquan (approved):

lib/worker.go
```diff
@@ -307,8 +305,8 @@ func (w *DefaultBaseWorker) MetaKVClient() metaclient.KVClient {
 	return w.userMetaKVClient
 }
 
-func (w *DefaultBaseWorker) UpdateStatus(ctx context.Context, status WorkerStatus) error {
-	err := w.statusSender.AsyncSendStatus(ctx, status)
+func (w *DefaultBaseWorker) UpdateStatus(ctx context.Context, status libModel.WorkerStatus) error {
+	err := w.statusSender.UpdateStatus(ctx, &status)
```
amyangfei, line 309: Please add comment about the rate limit and delivery guarantee of `UpdateStatus`.
If worker sends messages with higher frequency than the processing rate than master, the overflow messages will be dropped in master side, is it right

amyangfei (approved):

PR #259, opened 2022-03-28
DMJobmaster: add task manager

GMHDBJD:
- add task manager
- add task status


GMHDBJD:
cc @lance6716

jobmaster/dm/metadata/job.go
```diff
@@ -19,7 +19,9 @@ const (
 	StageRunning
 	StagePaused
 	StageFinished
-	StageDelete
+	// UnScheduled means the task is not scheduled.
+	// This usually happens when the worker is offline.
+	StageUnscheduled
```
lance6716, line 24: I'm not sure what's the difference of stage and status and other term. @liuzix could your give some advice on naming?

GMHDBJD: This stage only use in dm-master

jobmaster/dm/task_manager.go
```diff
@@ -0,0 +1,218 @@
...
+	tm.tasks.Range(func(key, value interface{}) bool {
+		result[key.(string)] = value.(runtime.TaskStatus)
+		return true
+	})
+	return result
+}
+
+// Trigger triggers the task manager to check and operate task.
+// TODO: Implement a more accurate and efficient way to trigger task manager if needed.
+func (tm *TaskManager) Trigger(ctx context.Context, delay time.Duration) {
```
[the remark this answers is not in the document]
GMHDBJD: update

jobmaster/dm/metadata/job.go
liuzix: `stage` seems to imply irreversibility, while `status` implies a state machine.

jobmaster/dm/scheduler/scheduler.go
```diff
@@ -0,0 +1,90 @@
...
+	s.lastCheckTime = now
+	s.nextCheckTime = now
+}
+
+func (s *DefaultScheduler) Trigger(delay time.Duration) {
+	t := time.Now().Add(delay)
+	s.setNextCheckTime(t)
+}
+
+func (s *DefaultScheduler) Run(ctx context.Context) {
```
liuzix, line 66: If this function is called directly by the job master, we can just reuse the `Tick` provided by the framework.

jobmaster/dm/metadata/job.go
lance6716: Will CDC also need this concept? we can find a better name at present.

jobmaster/dm/scheduler/scheduler.go
GMHDBJD: It called by job master indeed, But I think separating them reduces coupling.

GMHDBJD: Later it may add some heavy/slow operation, I think tick might be some lightweight operation.

jobmaster/dm/scheduler/scheduler.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (s *DefaultScheduler) advanceCheckTime() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	now := time.Now()
+	s.lastCheckTime = now
+	s.nextCheckTime = now
+}
+
+func (s *DefaultScheduler) Trigger(delay time.Duration) {
```
lance6716, line 61: "Trigger" is too ambiguous, can we directly use `setNextCheckTime`?

jobmaster/dm/scheduler/scheduler.go
lance6716: I think there're two ways to use `DefaultScheduler`

1.

```
func (j JobMaster) Tick() {
  j.scheduler.Tick()
}
```

2

```
func (j JobMaster) InitImpl() {
  go j.scheduler.Run()
}
```

And I remember @amyangfei says Tick can support some long-time running functions now. I prefer the first way, it's also get rid of implmenting out own tick loop.

jobmaster/dm/scheduler/scheduler.go
```diff
@@ -0,0 +1,90 @@
...
+	"time"
+
+	"github.com/pingcap/tiflow/dm/pkg/log"
+)
+
+const (
+	checkInterval = 50 * time.Millisecond
+)
+
+type Scheduler interface {
```
lance6716, line 15: Is this struct used for source worker scheduling or something? please add some comment

Or we can call it Ticker {Tick} to indicate its calling scenario.

jobmaster/dm/runtime/task_status.go
```diff
@@ -0,0 +1,111 @@
+package runtime
+
+import (
+	"encoding/json"
+
+	"github.com/hanfei1991/microcosm/jobmaster/dm/metadata"
+	"github.com/hanfei1991/microcosm/lib"
+	"github.com/pingcap/errors"
```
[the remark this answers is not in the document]
amyangfei: How do we define intra-project and inter-project, such as in this case, does 
- `github.com/pingcap/errors` is lib from infra-project
- and `github.com/hanfei1991/microcosm/lib` is lib from inter-project

lance6716 (approved):

jobmaster/dm/runtime/task_status.go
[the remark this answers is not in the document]
lance6716: I think he means 3-rd party imports and packages belong to this repo

GMHDBJD:
cc @liuzix @nongfushanquan

jobmaster/dm/ticker/ticker.go
```diff
@@ -0,0 +1,68 @@
...
+		errorInterval:  errorInterval,
+	}
+	defaultTicker.SetNextCheckTime(time.Now())
+	return defaultTicker
+}
+
+func (s *DefaultTicker) SetNextCheckTime(t time.Time) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if (t.After(s.lastCheckTime) && t.Before(s.nextCheckTime)) || s.lastCheckTime.Equal(s.nextCheckTime) {
```
amyangfei, line 37: It seems `t.After(s.lastCheckTime) && t.Before(s.nextCheckTime)` can't be triggered in code path, or is it better to check `s.lastCheckTime.Equal(s.nextCheckTime)` first to reduce one check.

amyangfei (approved):

PR #264, opened 2022-03-28
ansible: support etcd

hanfei1991:
Deploy etcd in where master runs

### DIFF
--- a/ansible/roles/master/templates/start.sh.j2
+++ b/ansible/roles/master/templates/start.sh.j2
@@ -14,4 +14,7 @@
     --advertise-addr {{ inventory_hostname }}:{{ master_port }} \
     --peer-urls 0.0.0.0:{{ master_peer_port }} \
     --advertise-peer-urls {{ inventory_hostname }}:{{ master_peer_port }} \
-    --initial-cluster {{ cluster|join(',') }}
+    --initial-cluster {{ cluster|join(',') }} \
+    --frame-meta-endpoints {{ inventory_hostname }}:{{ etcd_peer_port }} \
+    --user-meta-endpoints {{ inventory_hostname }}:{{ etcd_peer_port }}
+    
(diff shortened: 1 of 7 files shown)


ansible/roles/master/templates/start.sh.j2
```diff
@@ -14,4 +14,7 @@
     --advertise-addr {{ inventory_hostname }}:{{ master_port }} \
     --peer-urls 0.0.0.0:{{ master_peer_port }} \
     --advertise-peer-urls {{ inventory_hostname }}:{{ master_peer_port }} \
-    --initial-cluster {{ cluster|join(',') }}
+    --initial-cluster {{ cluster|join(',') }} \
+    --frame-meta-endpoints {{ inventory_hostname }}:{{ etcd_peer_port }} \
+    --user-meta-endpoints {{ inventory_hostname }}:{{ etcd_peer_port }}
```
amyangfei, line 19: Will there be any conflict when etct for framework and user are deployed in the same etcd? PTAL @maxshuang

amyangfei: I found both framework and user will write the key `test/test_key`

maxshuang: > Will there be any conflict when etct for framework and user are deployed in the same etcd? PTAL @maxshuang

No conflict. 
But user must don't use the same key as framework , because  framework and user use the same tenantID 'default-user' now.

amyangfei: Can change to use different tenantID, making it easy for debug (in the same etcd).

amyangfei (approved):
lgtm for the ansible deploy part.

ansible/roles/master/templates/start.sh.j2
maxshuang: > Can change to use different tenantID, making it easy for debug (in the same etcd).

Ok, I will make a pr for tenantID support later today.

PR #272, opened 2022-03-29
lib/master: refactor WorkerManager

liuzix:


lib/master/worker_entry.go
```diff
@@ -0,0 +1,99 @@
...
+func (e *workerEntry) MarkAsTombstone() {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	e.isTombstone = true
+}
+
+func (e *workerEntry) IsTombstone() bool {
+	return e.isTombstone
+}
```
amyangfei, line 67: Should access be protected by lock?

lib/master/worker_handle.go
```diff
@@ -0,0 +1,140 @@
...
+	if nonblocking {
+		_, err = h.manager.messageSender.SendToNode(ctx, p2p.NodeID(h.executorID), topic, message)
+	} else {
+		err = h.manager.messageSender.SendToNodeB(ctx, p2p.NodeID(h.executorID), topic, message)
+	}
+
+	if err != nil {
+		return err
+	}
+	return nil
```
amyangfei, line 102:
```suggestion
	return err
```

lib/master/worker_manager.go
```diff
@@ -0,0 +1,494 @@
...
+	timer := m.clock.Timer(timeoutInterval)
+	defer timer.Stop()
+
+	select {
+	case <-ctx.Done():
+		return errors.Trace(ctx.Err())
+	case <-timer.C:
+		// Wait for the worker timeout to expire, so
+		// that we maintain all the workers that are alive.
+	}
```
amyangfei, line 141: Does this mean master won't start to work until `m.timeouts.WorkerTimeoutDuration + m.timeouts.WorkerTimeoutGracefulDuration` (20s) passes

lib/master/worker_manager.go
```diff
@@ -0,0 +1,494 @@
...
+			}
+		case workerDispatched:
+			if err := m.onWorkerDispatched(ctx, event.Handle, event.Err); err != nil {
+				return err
+			}
+		}
+
+		if event.beforeHook != nil {
+			event.beforeHook()
+		}
```
amyangfei, line 258: Could you please explain when naming it `beforeHook`?

lib/master/worker_manager.go
```diff
@@ -0,0 +1,494 @@
...
+				manager:    m,
+			},
+			Err: errIn,
+		}
+	}
+	err := m.enqueueEvent(event)
+	if err != nil {
+		log.L().Warn("workerDispatchFailed event is dropped",
+			zap.Error(errIn))
+	}
```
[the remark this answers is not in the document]
liuzix: Lost notification is inevitable if there is a failover. But allowing lost notification at any time may burden the writers of the business too much.

lib/master/worker_manager.go
liuzix: It is a function called before the main callback is called.

lib/master/worker_manager.go
liuzix: Yes. We can change it so that if all workers recorded in meta have sent us a heartbeat, the master resumes working. But in cases where there is at least one worker crashing, the master still needs to wait for `20s`.

amyangfei (approved):

PR #282, opened 2022-03-31
ci: support dm e2e test in github action

amyangfei:
As title said.

### DIFF
--- a/sample/dm_databases.yaml
+++ b/sample/dm_databases.yaml
@@ -1,0 +1,49 @@
+version: '2.3'
+
+services:
+    dm_upstream_mysql:
+        image: mysql:5.7
+        container_name: dm_upstream_mysql
+        ports:
+            - "3306:3306"
+        volumes:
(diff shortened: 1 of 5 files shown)


sample/dm_databases.yaml
```diff
@@ -0,0 +1,49 @@
...
+        image: pingcap/tikv:v5.4.0
+        command:
+            - --addr=0.0.0.0:20160
+            - --advertise-addr=dm_downstream_tikv:20160
+            - --pd=dm_downstream_pd:2379
+        depends_on:
+            - "dm_downstream_pd"
+        restart: on-failure
+
+    dm_downstream_tidb:
```
lance6716, line 38: in fact we can use TiDB with unistore (no TiKV)

lance6716 (approved):

sample/dm_databases.yaml
amyangfei: Yep, I concern unistore is not a completed implementation with new features (or new feature is added later than TiKV), so I use TiKV here.

Issue #289, opened 2022-04-08
userMetaKVClient is not compatible with meta mock client


GMHDBJD:
### What did you do
```
kvClient := kvclient.NewPrefixKVClient(mock.NewMetaMock(), tenant.DefaultUserTenantID)
kvClient.Txn(ctx).Do(putOp).Commit()
```
### What did you see
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x15debb8]

goroutine 226 [running]:
testing.tRunner.func1.2({0x1de6d00, 0x3e48ce0})
	/usr/local/go/src/testing/testing.go:1389 +0x24e
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1392 +0x39f
panic({0x1de6d00, 0x3e48ce0})
	/usr/local/go/src/runtime/panic.go:838 +0x207
github.com/hanfei1991/microcosm/pkg/meta/namespace.(*kvPrefix).unprefixTxnResponse(0xc000f74a20?, 0x1?)
	/home/gmhdbjd/code/microcosm/pkg/meta/namespace/kv.go:149 +0x18
github.com/hanfei1991/microcosm/pkg/meta/namespace.(*txnPrefix).Commit(0xc000748b88)
	/home/gmhdbjd/code/microcosm/pkg/meta/namespace/kv.go:128 +0x53
github.com/hanfei1991/microcosm/jobmaster/dm/metadata.(*DefaultStore).Put(0xc000ca60f0, {0x25f8370, 0xc0000520b8}, {0x1da98a0?, 0xc000748a68})
	/home/gmhdbjd/code/microcosm/jobmaster/dm/metadata/state.go:79 +0x24f
github.com/hanfei1991/microcosm/jobmaster/dm/metadata.TestDefaultStore(0x2?)
	/home/gmhdbjd/code/microcosm/jobmaster/dm/metadata/state_test.go:69 +0x3b4
testing.tRunner(0xc000fbc680, 0x22109b8)
	/usr/local/go/src/testing/testing.go:1439 +0x102
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1486 +0x35f
FAIL	github.com/hanfei1991/microcosm/jobmaster/dm/metadata	0.174s
FAIL
```

PR #290, opened 2022-04-08
fix(mock): meta mock support txn and fix txn panic

maxshuang:
close  issue: #289 

meta mock support serialize level txn


GMHDBJD (approved):
LGTM

PR #291, opened 2022-04-11
jobmanager: delete master meta when create worker returns error

amyangfei:
Fix a problem that when create worker returns error during user submit job, the metadata of job master is still kept in metastore. And if the job manager transfers, the failed job master will be failover.
Just remove the master meta if a job is failed to create.

### DIFF
--- a/servermaster/jobmanager.go
+++ b/servermaster/jobmanager.go
@@ -154,6 +154,11 @@ func (jm *JobManagerImplV2) SubmitJob(ctx context.Context, req *pb.SubmitJobRequ
 	id, err = jm.BaseMaster.CreateWorker(
 		meta.Tp, meta, defaultJobMasterCost)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


servermaster/jobmanager.go
```diff
@@ -154,6 +154,11 @@ func (jm *JobManagerImplV2) SubmitJob(ctx context.Context, req *pb.SubmitJobRequ
 	id, err = jm.BaseMaster.CreateWorker(
 		meta.Tp, meta, defaultJobMasterCost)
 	if err != nil {
+		err2 := lib.DeleteMasterMeta(ctx, jm.BaseMaster.MetaKVClient(), meta.ID)
```
hanfei1991, line 157: If the delete op fails, do we need to gc it in the future?

hanfei1991 (approved):

PR #301, opened 2022-04-13
framework metaclient query support

maxshuang:
1. add the orm layer
2. add the access interface for framework metastore client


pkg/metautil/model.go
```diff
@@ -0,0 +1,56 @@
...
+	ProjectID string    `gorm:"type:char(64) not null;index:idx_op"`
+	Operation string    `gorm:"type:varchar(16) not null"`
+	JobID     string    `gorm:"type:char(64) not null"`
+	CreatedAt time.Time `gorm:"autoCreateTime;index:idx_op"`
+}
+
+type JobInfo struct {
+	Model
+	ProjectID string `gorm:"type:char(64) not null;index:idx_st,priority:1"`
+	JobID     string `gorm:"type:char(64) not null;uniqueIndex:uidx_id"`
```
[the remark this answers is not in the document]
maxshuang: Yes, it's the uuid the framework used. 
The precise length is 35 bytes, may be 36 is the most proper num.
Since the uuid is always the same length, so I prefer char instead of varchar.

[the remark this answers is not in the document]
amyangfei: So why choosing 64 instead of 36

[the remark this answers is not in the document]
maxshuang: Preserve some bytes in case of some change future.
But I will change to 36 for now.

pkg/orm/util.go
```diff
@@ -0,0 +1,364 @@
...
+	MaxOpenConns    int
+}
+
+// NewSqlDB return sql.DB for specified driver and dsn
+// dsn format: [username[:password]@][protocol[(address)]]/dbname[?param1=value1&...&paramN=valueN]
+// TODO: using projectID as dbname to construct dsn to achieve isolation
+func NewSQLDB(driver, dsn string, conf DBConfig) (*sql.DB, error) {
+	db, err := sql.Open(driver, dsn)
+	if err != nil {
+		return nil, cerrors.ErrMetaNewClientFail.GenWithStackByArgs(err)
```
amyangfei, line 45:
```suggestion
		return nil, cerrors.Wrap(cerrors.ErrMetaNewClientFail, err)
```

pkg/orm/util.go
```diff
@@ -0,0 +1,364 @@
...
+	cerrors "github.com/hanfei1991/microcosm/pkg/errors"
+	"github.com/hanfei1991/microcosm/pkg/orm/model"
+)
+
+// TODO: retry and idempotent??
+// TODO: context control??
+
+const (
+	DefaultConnMaxIdleTime = time.Duration(30)
+	DefaultConnMaxLifeTime = time.Duration(18600)
```
amyangfei, line 21: How is the default value `18600` chosen, it is 5h 10m.

pkg/orm/util.go
```diff
@@ -0,0 +1,364 @@
...
+	db.SetMaxIdleConns(conf.MaxIdleConns)
+	db.SetMaxOpenConns(conf.MaxOpenConns)
+
+	return db, nil
+}
+
+// NewMetaOpsClient return the client to operate framework metastore
+func NewMetaOpsClient(sqlDB *sql.DB) (*MetaOpsClient, error) {
+	if sqlDB == nil {
+		return nil, cerrors.ErrMetaOpFail.GenWithStackByArgs(fmt.Errorf("para is nil"))
```
amyangfei, line 59: please refine the generated error.

pkg/orm/util.go
```diff
@@ -0,0 +1,364 @@
...
+		return nil, cerrors.ErrMetaOpFail.GenWithStackByArgs(result.Error)
+	}
+
+	return &project, nil
+}
+
+// AddProjectOperation insert the operation
+func (c *MetaOpsClient) AddProjectOperation(ctx context.Context, op *model.ProjectOperation) error {
+	if op == nil {
+		return cerrors.ErrMetaOpFail.GenWithStackByArgs(fmt.Errorf("para is nil"))
```
amyangfei, line 140: ditto

pkg/orm/util.go
```diff
@@ -0,0 +1,364 @@
...
+	if result := c.db.Where("project_id = ? AND created_at >= ? AND created_at <= ?", projectID, tr.start,
+		tr.end).Find(&projectOps); result.Error != nil {
+		return nil, cerrors.ErrMetaOpFail.GenWithStackByArgs(result.Error)
+	}
+
+	return projectOps, nil
+}
+
+/////////////////////////////// Job Operation
+// SetJob insert the jobInfo
```
amyangfei, line 174:
```suggestion
// AddJob insert the jobInfo
```

pkg/orm/util.go
```diff
@@ -0,0 +1,364 @@
...
+// DeleteJob delete the specified jobInfo
+func (c *MetaOpsClient) DeleteJob(ctx context.Context, jobID string) error {
+	if result := c.db.Where("id = ?", jobID).Delete(&model.MasterMeta{}); result.Error != nil {
+		return cerrors.ErrMetaOpFail.GenWithStackByArgs(result.Error)
+	}
+
+	return nil
+}
+
+// QueryJobByID query job by `jobID`
```
amyangfei, line 206:
```suggestion
// GetJobByID query job by `jobID`
```

pkg/orm/util.go
```diff
@@ -0,0 +1,364 @@
...
+	if result := c.db.Where("id = ?", resourceID).
+		First(&resource); result.Error != nil {
+		return nil, cerrors.ErrMetaOpFail.GenWithStackByArgs(result.Error)
+	}
+
+	return &resource, nil
+}
+
+// QueryResourcesByJobID query all resources of the jobID
+func (c *MetaOpsClient) QueryResourcesByJobID(ctx context.Context, jobID string) ([]model.ResourceMeta, error) {
```
amyangfei, line 347: Better to unify the verb, use `GetXXX` or `QueryXXX`

pkg/orm/util.go
maxshuang: an enough long lifetime for the conn

amyangfei: - The unit of time.Duration is Nanosecond, these two default values are not seconds here
- Better to use a more meaningful default value to prevent other people to be confused

amyangfei (approved):

PR #303, opened 2022-04-14
dmJobmaster: add checkpoint agent

GMHDBJD:
- add checkpoint agent


GMHDBJD:
cc @lance6716

lance6716:
We should expose those functions in DM (Don't repeat yourself). If in future we'll change the SQL of checkpoint, we can only change one place.

GMHDBJD:
> We should expose those functions in DM (Don't repeat yourself). If in future we'll change the SQL of checkpoint, we can only change one place.
DM now create checkpoint in worker, this pr change it to master, I' will move the code back later after this pr merged(6.1.0 is LTS)

lance6716:
> > We should expose those functions in DM (Don't repeat yourself). If in future we'll change the SQL of checkpoint, we can only change one place.
> 
> DM now create checkpoint in worker, this pr change it to master. Many codes should be refactor. e.g. Units' init logic. I' will move the code back with those refactor.

Why we let DM-master/JobMaster create checkpoints? I think it's more decoupled that Unit/DMUnitWorker provide the IsFresh helper function (no need to construct an instance to call this function) and let subtaskHolder/JobMaster invoke it.

GMHDBJD:
https://github.com/hanfei1991/microcosm/blob/4ec424d64acf5d8e0f1c13e2e67dc78575f35eec/jobmaster/dm/checkpoint/checkpoint_agent.go#L74

GMHDBJD:
> I think it's more decoupled that Unit/DMUnitWorker provide the IsFresh helper function (no need to construct an instance to call this function) and let subtaskHolder/JobMaster invoke it.

I think IsFresh/Init/Remove are all db function, they should be a checkpoint PKG used by master and worker. I put it in master in this pr just for convenience, because it later will be move to tiflow repo.

lance6716:
> // We create checkpoint table in master rather than in workers,
> // to avoid the annoying log of "table already exists",

this log problem has many solutions, I think we should not be limited to let master create checkpoints.

> they should be a checkpoint PKG used by master and worker

I agree with it or we can put the package in worker package

> I put it in master pkg in this pr just for convenience

From my point of view, worker should cover all actions related to it, leaking the functionality will make the system more complex.

> If IsFresh func is in DMUnitWorker, we should create the unit first and then check it is fresh, it doesn't make sense.

We should avoid construct the unit, maybe we can register a map contains WorkerType -> func (taskName, ...) (bool error) and let master find the function by WorkerType

jobmaster/dm/checkpoint/agent.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/pingcap/tidb-tools/pkg/dbutil"
+	dmconfig "github.com/pingcap/tiflow/dm/dm/config"
+	"github.com/pingcap/tiflow/dm/pkg/conn"
+	"github.com/pingcap/tiflow/dm/pkg/cputil"
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"go.uber.org/zap"
+)
+
+const (
+	loadCheckpointTable = `CREATE TABLE IF NOT EXISTS ? (
```
lance6716, line 21: I'm not sure this `?` can be used to represent identifiers :thinking: 

> Parameter markers can be used only where data values should appear, not for SQL keywords, identifiers, and so forth.

https://dev.mysql.com/doc/refman/5.7/en/prepare.html

jobmaster/dm/checkpoint/agent.go
```diff
@@ -0,0 +1,206 @@
...
+func shardMetaName(cfg *config.JobCfg) string {
+	return dbutil.TableName(cfg.MetaSchema, cputil.SyncerShardMeta(cfg.Name))
+}
+
+func onlineDDLName(cfg *config.JobCfg) string {
+	return dbutil.TableName(cfg.MetaSchema, cputil.SyncerOnlineDDL(cfg.Name))
+}
+
+func isLoadFresh(ctx context.Context, taskCfg *config.TaskCfg, db *conn.BaseDB) (bool, error) {
+	query := "SELECT status FROM ? WHERE `task_name` = ? AND `source_name` = ?"
```
lance6716, line 181:
```suggestion
	query := "SELECT status FROM %s WHERE `task_name` = ? AND `source_name` = ?"
```

lance6716 (commented):
rest lgtm

lance6716 (approved):

GMHDBJD:
cc @amyangfei @liuzix @nongfushanquan

lance6716:
UT failed.

It's a good chance that we fix unstable tests when the project is small!

GMHDBJD:
> It's a good chance that we fix unstable tests when the project is small!
I'll try to fix it.

amyangfei (approved):

amyangfei:
> UT failed.
> 
> It's a good chance that we fix unstable tests when the project is small!

Good suggestion, we have tracking these unstable tests in https://github.com/hanfei1991/microcosm/issues/313

PR #310, opened 2022-04-18
lib/master: implement and use ErrCenter

liuzix:
- Explore a more managed mechanism of error handling inside `BaseMaster` and `BaseWorker`.

Abstractly speaking, this model is like this: there is an object O, and there are methods A, B, C, D on it. These methods may be called by many goroutines. These methods may depend on other modules or third-party libraries, and the dependent methods may block. Suppose an error  occurs in a method of object O, and we want to stop all methods of O as soon as possible, and call it again with an error.


pkg/errctx/center.go
```diff
@@ -0,0 +1,40 @@
+package errctx
+
+import (
+	"context"
+
+	"go.uber.org/atomic"
+)
+
+type ErrCenter struct {
```
lance6716, line 9: prefer the name `FirstError`

pkg/errctx/center.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	c.errVal.Store(err)
+	close(c.doneCh)
+}
+
+func (c *ErrCenter) CheckError() error {
+	return c.errVal.Load()
+}
+
+func (c *ErrCenter) DeriveContext(ctx context.Context) context.Context {
```
lance6716, line 38: we can make it more like a standard library stype, WithCancelOnError(ctx, firstErr) or something

pkg/errctx/ctx.go
```diff
@@ -0,0 +1,67 @@
...
+	// has true.
+	c.doneCh = doneCh
+	c.hasDoneCh.Store(true)
+
+	return doneCh
+}
+
+func (c *errCtx) Err() error {
+	if c.center.hasErr.Load() {
+		return c.center.errVal.Load()
```
lance6716, line 63: there's a time slice in setting `hasErr` and `errVal` in `OnError`

liuzix: Good catch. I will use an RW-lock instead.

pkg/errctx/center.go
liuzix: I think `FirstError` is too simple to describe the object's functionality. Don't you think so?

lance6716: From `ErrCenter` I can not know its functionality, maybe left some comments

pkg/errctx/center.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (c *ErrCenter) OnError(err error) {
+	if err == nil {
+		return
+	}
+	if c.hasErr.Swap(true) {
+		// OnError is no-op after the first call with
+		// a non-nil error.
+		return
+	}
```
amyangfei, line 29: If some of scenarios need to record the duplicated errors, should we provide an option to do that.

pkg/errctx/ctx.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (c *errCtx) Done() <-chan struct{} {
+	if c.hasDoneCh.Load() {
+		// Fast path. No lock contention.
+		return c.doneCh
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
maxshuang, line 33: sync.Once may be a better choice ??

pkg/errctx/center.go
liuzix: Allowing duplicate errors can make this utility too difficult to use. I added a log instead.

pkg/errctx/ctx.go
```diff
@@ -0,0 +1,50 @@
...
+			case <-c.center.doneCh:
+			case <-c.Context.Done():
+			}
+
+			close(doneCh)
+		}()
+
+		// There is no data race here because
+		// c.doneCh is read from only if c.hasDoneCh
+		// has true.
```
amyangfei, line 38: The comment is outdated

amyangfei (commented):
rest lgtm

amyangfei (approved):

PR #317, opened 2022-04-19
lib/worker: use task submit time as worker's first heartbeat timeout baseline

liuzix:
- Fix the problem if the runtime delayed too much before `Init`, the worker will not commit suicide in time.


lib/worker.go
```diff
@@ -199,12 +197,20 @@ func (w *DefaultBaseWorker) doPreInit(ctx context.Context) error {
 			zap.Error(err))
 	}()
 
+	w.startBackgroundTasks()
+
+	initTime := w.clock.Mono()
+	rctx, ok := runtime.ToRuntimeCtx(ctx)
+	if ok {
+		initTime = clock.ToMono(rctx.SubmitTime())
```
amyangfei, line 205: The submit time in the context is set in task runner, will it be adjusted to the real time  of `create worker` in master. 

https://github.com/hanfei1991/microcosm/blob/6f1e01c87b8f7feecd777b43645ef80875b45636/executor/worker/task_runner.go#L78

amyangfei: I didn't describe `create worker returns` clearly, given the following logic
 
https://github.com/hanfei1991/microcosm/blob/6f1e01c87b8f7feecd777b43645ef80875b45636/lib/master.go#L533-L559

- worker is added to worker manager in L533, which is before `t0` strictly, assuming the two machines have the same wall time. let's say it is `t2`. The heartbeat timeout counter works since worker is added into worker manager
- `t0` happens after L542 runs
- Master treats worker timeout at `t2+delta`, meantime worker doesn't think it is timeout, because the passed time is `t2+delta-t0`, which is less than `delta`

liuzix: Okay I see the problem here. Looks like some design of the `WorkerManager` needs to be adjusted. I still think exchange times between machines can be problematic.

amyangfei (approved):

PR #318, opened 2022-04-19
make demo test scalable

hanfei1991:
1. support for deploying multiple demo servers
2. demo servers can use memory db


jobmaster/cvsJob/cvsJobMaster.go
```diff
@@ -383,19 +395,21 @@ func (jm *JobMaster) IsJobMasterImpl() {
...
+	//client := pb.NewDataRWServiceClient(conn)
+	//defer conn.Close()
+	//reply, err := client.ListFiles(ctx, &pb.ListFilesReq{})
+	//if err != nil {
+	//	log.L().Info(" list the directory failed ", zap.String("id", jm.ID()), zap.Error(err))
+	//	return -1, err
+	//}
+	//return int(reply.FileNum), nil
+
+	return 50, nil
```
amyangfei, line 412: Why returning a hard code count here instead of using `ListFiles` API.

amyangfei (approved):

PR #321, opened 2022-04-20
servermaster: set initialized to true after leader runs

amyangfei:
This PR fixes several bugs in server master and worker manger

- Set server master `leaderInitialized` after run leader service runs
- Pre allocate response struct before calling `PreRPC` to avoid assigning to an empty ptr.
- Remove lock acquiring in workerEntry.String, because it can be accessed in a locked path, which will cause deadlock, such as L136 and L 144, https://github.com/hanfei1991/microcosm/blob/5c44fa8091ead367a0ee1f1dd52d9f9d56e6659a/lib/master/worker_entry.go#L135-L145
- Another deadlock in worker manger, which is caused by panic in inner path, and lock is not released in outer path, so extract a `checkWorkerEntriesOnce` function.
- Refine the fast path for no worker logic during worker manger failover, it is a temporary fix for finished worker, we need to refine logic of worker manager failover with different kinds of worker.

### DIFF
--- a/pkg/rpcutil/server.go
+++ b/pkg/rpcutil/server.go
@@ -146,7 +146,8 @@ func (h PreRPCHook[T]) checkInitialized(respPointer interface{}) (shouldRet bool
 		errInRefelct := reflect.ValueOf(&pb.Error{
 			Code: pb.ErrorCode_MasterNotReady,
 		})
-		reflect.ValueOf(respPointer).Elem().FieldByName("Err").Set(errInRefelct)
+		reflect.Indirect(reflect.ValueOf(respPointer)).Elem().
+			FieldByName("Err").Set(errInRefelct)
 		return true
 	}
 	return false
--- a/servermaster/server.go
+++ b/servermaster/server.go
@@ -86,7 +86,7 @@ type Server struct {
 
 	metaStoreManager MetaStoreManager
 
-	initialized atomic.Bool
+	leaderInitialized atomic.Bool
 
 	// mocked server for test
 	mockGrpcServer mock.GrpcServer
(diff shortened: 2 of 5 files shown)


pkg/rpcutil/server.go
```diff
@@ -146,7 +146,8 @@ func (h PreRPCHook[T]) checkInitialized(respPointer interface{}) (shouldRet bool
 		errInRefelct := reflect.ValueOf(&pb.Error{
 			Code: pb.ErrorCode_MasterNotReady,
 		})
-		reflect.ValueOf(respPointer).Elem().FieldByName("Err").Set(errInRefelct)
```
lance6716, line 149: I should add unit tests in my PR 😂

servermaster/server.go
```diff
@@ -686,6 +686,7 @@ func (s *Server) runLeaderService(ctx context.Context) (err error) {
 			log.L().Warn("job manager close with error", zap.Error(err))
 		}
 	}()
+	s.leaderInitialized.Store(true)
```
lance6716, line 689: I think we can remove the time.Sleep(time.Second) in e2e_dm_test.go now

lance6716 (approved):

lance6716:
I forget to check the `respPointer` has Err member for `*rpcutil.PreRPCHook[pb.ResourceManagerClient]`. I'll submit a PR after this PR is merged.

lance6716:
> Another deadlock in worker manger, which is caused by panic in inner path

What's the reason of the panic?

amyangfei:
> > Another deadlock in worker manger, which is caused by panic in inner path
> 
> What's the reason of the panic?

panic reason is tracked in https://github.com/hanfei1991/microcosm/issues/322

servermaster/server.go
amyangfei: Yep, use require.Eventually() now.

PR #324, opened 2022-04-21
servermaster: integrate the refactored scheduler with Server

liuzix:
- Integrate the refactored scheduler with servermaster.
- No new feature is supported in this PR. Supporting scheduling by resources will be added together with other proto changes.


servermaster/resource/manager.go
```diff
@@ -3,17 +3,23 @@ package resource
...
 	// Register registers new executor, it is called when an executor joins
 	Register(id model.ExecutorID, addr string, capacity model.RescUnit)
 
 	// Unregister is called when an executor exits
 	Unregister(id model.ExecutorID)
 
-	// Allocate allocates executor resources to given tasks
+	// Allocate allocates executor resources to given tasks.
+	// Deprecated.
 	Allocate(tasks []*pb.ScheduleTask) (bool, *pb.TaskSchedulerResponse)
```
amyangfei, line 23: Can we remove the deprecated API now?

servermaster/scheduler/scheduler.go
```diff
@@ -71,7 +71,11 @@ func (s *Scheduler) checkCostAllows(
 	request *schedModel.SchedulerRequest,
 	target model.ExecutorID,
 ) bool {
-	executorResc := s.capacityProvider.CapacityForExecutor(target)
+	executorResc, ok := s.capacityProvider.CapacityForExecutor(target)
```
amyangfei, line 74: The `CapacityForExecutor` or `CapacitiesForAllExecutors` returns a capacity snapshot, if the `scheduler.ScheduleTask` is called concurrently, the capacity usage could have write skew, such as available capacity is 100, and two concurrent tasks asks for 80 cap each other, they are both allowed to start.

Is there any concurrency control to avoid such problem.

liuzix: Added comments on `CapacityProvider`.

servermaster/resource/manager.go
liuzix: Done.

amyangfei (approved):

PR #327, opened 2022-04-21
Add dm e2e test

GMHDBJD:
- add e2e test for with dm jobmaster


jobmaster/dm/checkpoint/agent.go
```diff
@@ -178,9 +188,10 @@ func onlineDDLName(cfg *config.JobCfg) string {
 }
 
 func isLoadFresh(ctx context.Context, taskCfg *config.TaskCfg, db *conn.BaseDB) (bool, error) {
-	query := "SELECT status FROM %s WHERE `task_name` = ? AND `source_name` = ?"
+	// nolint:gosec
+	query := fmt.Sprintf("SELECT status FROM %s WHERE `task_name` = ? AND `source_name` = ?", loadTableName((*config.JobCfg)(taskCfg)))
```
GMHDBJD, line 192: Both %s and ? tasks no effect for tablename, so use fmt instead.

GMHDBJD:
cc @lance6716

dm/unit.go
```diff
@@ -77,19 +87,42 @@ func (u *unitHolder) getResult() (bool, *pb.ProcessResult) {
...
+		statusBytes, err := json.Marshal(status)
+		if err != nil {
+			return err
+		}
 		s := libModel.WorkerStatus{
-			Code: libModel.WorkerStatusNormal,
+			Code:     libModel.WorkerStatusNormal,
+			ExtBytes: statusBytes,
 		}
 		// nolint:errcheck
```
lance6716, line 111:
```suggestion
```

dm/unit.go
```diff
@@ -116,24 +151,44 @@ func (u *unitHolder) tryUpdateStatus(ctx context.Context, base lib.BaseWorker) e
...
 			Code:         libModel.WorkerStatusError,
 			ErrorMessage: unit.JoinProcessErrors(result.Errors),
+			ExtBytes:     statusBytes,
 		}
 		// TODO: UpdateStatus too frequently?
 		// nolint:errcheck
 		_ = base.UpdateStatus(ctx, s)
 		return nil
 	case worker.ResumeDispatch:
+		u.unit.Pause()
```
lance6716, line 176: maybe we can check if the unit has been paused by ResumeSkip. But it should also be OK that we Pause twice

jobmaster/dm/checkpoint/agent.go
```diff
@@ -130,13 +134,19 @@ func (c *AgentImpl) IsFresh(ctx context.Context, workerType lib.WorkerType, task
 	return isSyncFresh(ctx, task.Cfg, db)
 }
 
+func createMetaDatabase(ctx context.Context, cfg *config.JobCfg, db *conn.BaseDB) error {
+	query := fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s", cfg.MetaSchema)
```
lance6716, line 138: should backquote it. we can use tiflow/pkg/quotes.QuoteName or dbutil like below

jobmaster/dm/message_agent.go
```diff
@@ -71,7 +71,9 @@ func (agent *MessageAgent) CreateWorker(ctx context.Context, taskID string, work
 	if _, ok := agent.sendHandles.Load(taskID); ok {
 		return "", errors.Errorf("worker for task %s already exist", taskID)
 	}
-	return agent.master.CreateWorker(workerType, taskCfg, 1)
+	// TODO: deprecated subtask cfg
```
lance6716, line 74: what's the plan for this?

lance6716 (approved):

jobmaster/dm/message_agent.go
GMHDBJD: May not

GMHDBJD:
cc @amyangfei @liuzix

amyangfei (approved):

PR #328, opened 2022-04-21
runtime: implement TaskCommitter

liuzix:
- Part I of implementing two-phase create worker.

### DIFF
--- a/executor/worker/task_committer.go
+++ b/executor/worker/task_committer.go
@@ -1,0 +1,171 @@
+package worker
+
+import (
+	"sync"
+	"time"
+
+	"github.com/pingcap/tiflow/dm/pkg/log"
+	"go.uber.org/zap"
+
+	"github.com/hanfei1991/microcosm/executor/worker/internal"
+	"github.com/hanfei1991/microcosm/pkg/clock"
+)
+
+const runTTLCheckerInterval = 1 * time.Second
+
+type requestID = string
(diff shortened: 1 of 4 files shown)


executor/worker/task_committer.go
```diff
@@ -0,0 +1,203 @@
...
+
+		c.removeRequestByID(request.RequestID)
+	}
+
+	// We use the current time as the submit time of the task.
+	taskWithSubmitTime := internal.WrapRunnable(task, c.clock.Now())
+
+	request := &requestEntry{
+		RequestID: rID,
+		ExpireAt:  c.clock.Now().Add(c.requestTTL),
```
maxshuang, line 110: Do we have different expectation of the requestTTL and taskWithSubmitTime?
In detail, should requestTTL be shorter than the worker timeout?

maxshuang:
others LGTM.

executor/worker/task_committer.go
liuzix: It does not matter. The client should not know about `requestTTL`, and if it wants to recreate a worker with the same name, the old request will be rewritten.

liuzix: We trust that the worker can cancel itself if it has timed out. We do not need another mechanism to prevent a worker from being run.

amyangfei (approved):

PR #344, opened 2022-04-25
*: Implement Two-Phase CreateWorker (Part II)

liuzix:
- Updated `executor.proto`.
- Implemented `PreDispatchTask` and `ConfirmTask` in the executor using `TaskCommitter`.
- Implemented `TaskDispatcher` for the `BaseMaster` to use.
- The implementation of `CreateWorker` is still **unchanged**, which will be done in another PR.


client/task_dispatcher.go
```diff
@@ -0,0 +1,212 @@
...
+		// NOTE: Aborted and AlreadyExists are guaranteed NOT to be
+		// generated by the gRPC framework.
+		// Refer to https://pkg.go.dev/google.golang.org/grpc/codes
+		case codes.Aborted:
+			// The business logic should be notified.
+			return "", false, errors.Trace(err)
+		case codes.AlreadyExists:
+			// Since we are generating unique UUIDs, this should not happen.
+			log.L().Panic("Unexpected error", zap.Error(err))
+			panic("unreachable")
```
amyangfei, line 172:
```suggestion
```

client/task_dispatcher.go
```diff
@@ -0,0 +1,212 @@
...
+	ctx, cancel := context.WithTimeout(ctx, d.timeout)
+	defer cancel()
+
+	requestID, err := d.preDispatchTaskWithRetry(ctx, args)
+	if err != nil {
+		return derrors.ErrExecutorPreDispatchFailed.Wrap(err)
+	}
+
+	// The timer must be started before invoking ConfirmDispatchTask.
+	startWorkerTimer()
```
amyangfei, line 75: Is this timer used for checking worker timeout after `confirmDispatchTask` returns success or returns undetermined error. Why it is must be started before invoking `confirmDispatchTask`

liuzix: - The timer is a virtual concept. `startWorkerTimer` actually corresponds to creating a `workerEntry` with an `ExpireAt` counting from now.
- It must be called before invoking `confirmDispatchTask` is because the Runtime will start the worker once it receives `ConfirmDispatchTask`. So we are expect heartbeats soon after calling `ConfirmDispatchTask`. To make sure no heartbeat is missed, or other unexpected problems, we should call `startWorkerTimer` before invoking `ConfirmDispatchTask`.

amyangfei (approved):

PR #346, opened 2022-04-26
*: Adjust the implementation of CreateWorker

liuzix:
- This is the Part III of implementing two-phase task dispatching.
- Reorganized `CreateWorker`.
- Adjusted the interface of `WorkerManager`.
- `OnWorkerDispatched` is now only called when dispatching worker has failed.


lib/master.go
```diff
@@ -522,56 +522,53 @@ func (m *DefaultBaseMaster) CreateWorker(
...
 		})
+
 		if err != nil {
-			// The executor may have already launched the worker.
-			// TODO summarize the kind of errors that could be received
-			// after success.
-			m.workerManager.OnCreatingWorkerFinished(workerID, nil)
+			log.L().Info("DispatchTask failed",
+				zap.Any("args", dispatchArgs),
+				zap.Error(err))
```
amyangfei, line 566: Better to add a comment that all clear work has been done in `AbortWorkerCallback`, and log error here is enough.

lib/master/events.go
```diff
@@ -8,7 +8,7 @@ const (
 	workerOnlineEvent = masterEventType(iota + 1)
 	workerOfflineEvent
 	workerStatusUpdatedEvent
-	workerDispatched
+	workerDispatchedEvent
```
amyangfei, line 11: Is `workerDispatchFailedEvent` more accurate?

amyangfei (approved):

PR #354, opened 2022-04-28
feat(orm): add more function to support framework client(part II)

maxshuang:
enhance orm client to support framework metaclient replacement


pkg/orm/client.go
```diff
@@ -0,0 +1,610 @@
...
+	end   time.Time
+}
+
+type Client interface {
+	metaclient.Client
+
+	// Initialize will create all tables for backend operation
+	Initialize(ctx context.Context) error
+
+	// project
```
liuzix, line 36: Can you group these methods into several interfaces?

pkg/orm/client.go
```diff
@@ -0,0 +1,610 @@
...
+	AddProject(ctx context.Context, project *model.ProjectInfo) error
```
liuzix, line 37:
```suggestion
	CreateProject(ctx context.Context, project *model.ProjectInfo) error
```

pkg/orm/client.go
```diff
@@ -0,0 +1,610 @@
...
+	DeleteProject(ctx context.Context, projectID string) error
+	QueryProjects(ctx context.Context) ([]*model.ProjectInfo, error)
+	GetProjectByID(ctx context.Context, projectID string) (*model.ProjectInfo, error)
+
+	// project operation
+	AddProjectOperation(ctx context.Context, op *model.ProjectOperation) error
```
liuzix, line 43:
```suggestion
	CreateProjectOperation(ctx context.Context, op *model.ProjectOperation) error
```

pkg/orm/client.go
```diff
@@ -0,0 +1,610 @@
...
+	QueryProjectOperations(ctx context.Context, projectID string) ([]*model.ProjectOperation, error)
```
liuzix, line 44: Do we need pagination and cursor here?

pkg/orm/client.go
```diff
@@ -0,0 +1,610 @@
...
+	QueryProjectOperationsByTimeRange(ctx context.Context, projectID string, tr TimeRange) ([]*model.ProjectOperation, error)
```
liuzix, line 45: Do we need a method with `order by time desc limit N`?

pkg/orm/client.go
```diff
@@ -0,0 +1,610 @@
...
+
+	// job info
+	UpsertJob(ctx context.Context, job *libModel.MasterMetaKVData) error
```
liuzix, line 48: With an RDBMS as a storage, do we still need the upsert method?

pkg/orm/client.go
```diff
@@ -0,0 +1,610 @@
...
+	UpdateJob(ctx context.Context, job *libModel.MasterMetaKVData) error
+	DeleteJob(ctx context.Context, jobID string) error
+	GetJobByID(ctx context.Context, jobID string) (*libModel.MasterMetaKVData, error)
+	QueryJobs(ctx context.Context) ([]*libModel.MasterMetaKVData, error)
+	QueryJobsByProjectID(ctx context.Context, projectID string) ([]*libModel.MasterMetaKVData, error)
+	QueryJobsByStatus(ctx context.Context, jobID string, status int) ([]*libModel.MasterMetaKVData, error)
+
+	// worker status
+	UpsertWorker(ctx context.Context, worker *libModel.WorkerStatus) error
```
liuzix, line 57: ditto

pkg/orm/client.go
```diff
@@ -0,0 +1,610 @@
...
+// metaOpsClient is the meta operations client for framework metastore
+type metaOpsClient struct {
+	// gorm claim to be thread safe
+	db   *gorm.DB
+	impl *sql.DB
+}
+
+func (c *metaOpsClient) Close() error {
+	if c.impl != nil {
+		return c.impl.Close()
```
liuzix, line 190: `errors.Trace()` or wrapping the error?

pkg/orm/client.go
```diff
@@ -0,0 +1,610 @@
...
+		if err := tx.First(&logicEp, defaultEpochPK).Error; err != nil {
+			return err
+		}
+		epoch = libModel.Epoch(logicEp.Epoch)
+
+		// return nil will commit the whole transaction
+		return nil
+	})
+	if err != nil {
+		return libModel.Epoch(0), cerrors.ErrMetaOpFail.Wrap(err)
```
liuzix, line 259: Retry here?

pkg/orm/client.go
maxshuang: We can add more interface here if we need.
But not in this pr.

pkg/orm/client.go
maxshuang: It's for the easy use of framework metaclient replacement.

When we insert the job info/worker status/resource meta after normal startup or failover, we need to check the existence of the item first, then do insert or update if we don't have the `Upsert` interface.

So here I introduce `Upsert`. And I will refine the implement using `replace into` later.

pkg/orm/client.go
maxshuang: I want the retry logic be a upper layer.
It will do it later.

pkg/orm/client.go
maxshuang: Thanks. I will add a TODO here and do it later.

pkg/orm/model/logic_epoch.go
```diff
@@ -0,0 +1,7 @@
+package model
+
+// using transaction to generate increasing epoch
+type LogicEpoch struct {
+	Model
+	Epoch int `gorm:"type:bigint not null default 1"`
```
amyangfei, line 6: Is `int64` better?

pkg/orm/client.go
```diff
@@ -0,0 +1,629 @@
...
+	}
+
+	if err := c.db.Create(job).Error; err != nil {
+		if !isDuplicateEntryErr(err) {
+			return cerrors.ErrMetaOpFail.Wrap(err)
+		}
+		if err := c.UpdateJob(ctx, job); err != nil {
+			return err
+		}
+	}
```
amyangfei, line 379: Can we use the [upsert on conflict](https://gorm.io/docs/create.html#Upsert-On-Conflict) feature that is provided in gorm. Ditto for other upsert implement.

pkg/orm/client.go
```diff
@@ -0,0 +1,612 @@
...
+	UpsertResource(ctx context.Context, resource *resourcemeta.ResourceMeta) error
+	UpdateResource(ctx context.Context, resource *resourcemeta.ResourceMeta) error
+	DeleteResource(ctx context.Context, resourceID string) error
+	GetResourceByID(ctx context.Context, resourceID string) (*resourcemeta.ResourceMeta, error)
+	QueryResources(ctx context.Context) ([]*resourcemeta.ResourceMeta, error)
+	QueryResourcesByJobID(ctx context.Context, jobID string) ([]*resourcemeta.ResourceMeta, error)
+	QueryResourcesByExecutorID(ctx context.Context, executorID string) ([]*resourcemeta.ResourceMeta, error)
+}
+
+// NewMetaOpsClient return the client to operate framework metastore
```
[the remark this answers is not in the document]
maxshuang: I want to concat the ProjectID into the dsn.
Should we use the default db `test` without projectID?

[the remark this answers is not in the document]
amyangfei: Is this `NewClient` used for framework only, if so `tenant.ProjectID` may be not necessary.

pkg/orm/client.go
maxshuang: I try the `upsert`. 
It seems more proper implement for this interface.
thanks.

pkg/externalresource/resourcemeta/model.go
```diff
@@ -46,6 +46,30 @@ func (m *ResourceMeta) ToQueryResourceResponse() *pb.QueryResourceResponse {
 	}
 }
 
+// Columns is used for updating the orm model
+func (m *ResourceMeta) Columns() []string {
```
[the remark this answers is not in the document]
maxshuang: Good suggestion. 
It's used in `Updatexxx` interface

pkg/orm/client.go
```diff
@@ -0,0 +1,614 @@
...
+	// we don't use `Save` here to avoid user dealing with the basic model
+	if err := c.db.Model(&resourcemeta.ResourceMeta{}).Where("id = ?", resource.ID).Updates(resource.Map()).Error; err != nil {
+		return cerrors.ErrMetaOpFail.Wrap(err)
+	}
+
+	return nil
+}
+
+// DeleteResource delete the specified model.libModel.resourcemeta.ResourceMeta
+func (c *metaOpsClient) DeleteResource(ctx context.Context, resourceID string) error {
```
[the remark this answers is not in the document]
maxshuang: Get.

pkg/orm/model/common.go
```diff
@@ -5,8 +5,9 @@ import (
 )
 
 // CreatedAt/UpdatedAt will autoupdate in the gorm lib, not in sql backend
+// TODO: refine the updatedAt to autoupdate backend, make `upsert` interface more elegant
 type Model struct {
-	SeqID     uint `gorm:"primaryKey;autoIncrement"`
-	CreatedAt time.Time
-	UpdatedAt time.Time
+	SeqID     uint      `json:"seq-id" gorm:"primaryKey;autoIncrement"`
```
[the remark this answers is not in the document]
maxshuang: basic model for the user's model.
Be the primary key for every item.
I can be removed, but I think it's normal to have a readable index for every item.

pkg/externalresource/resourcemeta/model.go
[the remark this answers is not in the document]
maxshuang: It is related to some gorm inner implement. 
I add a TODO here and do it later.
In this pr, I will move it to a variable to avoid useless object allocate.

amyangfei (commented):
rest lgtm

amyangfei (approved):

PR #361, opened 2022-05-05
lib/*(engine): Implement worker exit fast path

### LINKED ISSUE #339: Fast Path for OnWorkerOffline when worker graceful exit
It tasks 20s when jobmaster receive worker offline after worker finished now.
```
[2022/04/24 09:41:38.200 +00:00] [INFO] [task_runner.go:228] ["Task stopped"] [id=1d51c99d-18a1-45aa-8aa1-4aa100cfbe4f] [error="[DFLOW:ErrWorkerFinish]worker finished and exited"]
[2022/04/24 09:41:58.596 +00:00] [INFO] [dm_jobmaster.go:151] ["on worker finished"] [id=dc954673-2530-479e-83a9-c5ac64640e00] [worker_id=1d51c99d-18a1-45aa-8aa1-4aa100cfbe4f]
```

liuzix:
close #339 
- Still experimenting. Does not need merging.


lib/worker.go
```diff
@@ -269,11 +282,31 @@ func (w *DefaultBaseWorker) doPostInit(ctx context.Context) error {
 }
 
 func (w *DefaultBaseWorker) doPoll(ctx context.Context) error {
-	if err := w.messageHandlerManager.CheckError(ctx); err != nil {
-		return err
+	if err := w.errCenter.CheckError(); err != nil {
```
maxshuang, line 285: I notice that ErrCenter only cares about the first error.
What if we get a app error first here and trigger the suicide next, does it that we can't handle suicide?

lib/master/worker_entry.go
```diff
@@ -53,6 +54,8 @@ type workerEntry struct {
 	expireAt time.Time
 	state    workerEntryState
 
+	receivedFinish atomic.Bool
```
maxshuang, line 57: Finished is not a state for worker? 
why we can't use `workEntryFinished`?

lib/worker.go
```diff
@@ -269,11 +282,31 @@ func (w *DefaultBaseWorker) doPostInit(ctx context.Context) error {
...
+			w.workerExitFsm.Store(workerExited)
+			return err
+		}
+
+		switch w.workerExitFsm.Load() {
+		case workerNormal:
+			w.workerExitFsm.Store(workerHalfExit)
+			return derror.ErrWorkerHalfExit.FastGenByArgs()
+		case workerHalfExit:
+			if w.masterClient.IsMasterSideClosed() {
```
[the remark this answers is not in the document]
liuzix: It would mean a bug in `BaseMaster`. But you are right, we need to have a mechanism to handle it gracefully.

sample/config/etcd.yml
```diff
@@ -135,6 +135,3 @@ log-outputs: [stderr]
 
 # Force to create a new one member cluster.
 force-new-cluster: false
-
-auto-compaction-mode: revision
```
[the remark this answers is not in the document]
liuzix: Because some integration test cases depend on Etcd Watch. Lost revisions would make the test fail.

lib/master/worker_entry.go
liuzix: `Finished` is technically a second component of the state. Formally, introducing the `receivedFinish` field makes the FSM states a Cartesian product `(workerEntryState, isFinished)`. For example, we can have `(workerEntryNormal, true)` or `(workerEntryOffline, false)`. `receivedFinish` is similar to a `signal-received` flag, which needs to be handled in parallel to its other states.

lib/worker.go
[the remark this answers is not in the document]
maxshuang: IsMasterSideClosed is set by the heartbeat pong.
So when we lost the connection at that time, it's possible we can get the workerExit state.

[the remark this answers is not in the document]
liuzix: Resolved in the latest commit

lib/worker.go
liuzix: Resolved in the latest commit

lib/worker.go
```diff
@@ -672,9 +707,97 @@ func (m *masterClient) SendHeartBeat(ctx context.Context, clock clock.Clock) err
...
+// PollExit is called in each tick of the worker.
+// Returning an error other than ErrWorkerHalfExit
+// means that the worker is ready to exit.
+func (c *workerExitController) PollExit() error {
+	err := c.errCenter.CheckError()
+	if err == nil {
+		return nil
+	}
+
+	if derror.ErrWorkerSuicide.Equal(err) {
```
maxshuang, line 760: Seems this logic is unneccesary.

lib/worker.go
```diff
@@ -672,9 +707,97 @@ func (m *masterClient) SendHeartBeat(ctx context.Context, clock clock.Clock) err
...
+		c.workerExitFsm.Store(workerHalfExit)
+		c.halfExitTime.Store(c.clock.Now())
+		return derror.ErrWorkerHalfExit.FastGenByArgs()
+	case workerHalfExit:
+		if c.masterClient.IsMasterSideClosed() {
+			c.workerExitFsm.Store(workerExited)
+			return err
+		}
+		sinceStartExiting := c.clock.Since(c.halfExitTime.Load())
+		if sinceStartExiting > workerExitWaitForMasterTimeout {
```
maxshuang, line 777: I have a concern for this timeout(15s) here.
We have the guarantee previously that master need recreate a new worker before worker commit suicide.
Do you have other assert to check that guarantee? Or we will need to wait at most WorkerTimeoutDuration+ workerExitWaitForMasterTimeout (15s+15s) if workerExitFsm(Exit) is overrided by  PollExit logic.

liuzix: > We have the guarantee previously that master need recreate a new worker before worker commit suicide.

The master can recreate the worker **after** the worker commits suicide. It is the master's side that needs to wait longer to ensure safety.

maxshuang: My mistake for the wrong statement. It's `After`.
We need some assertion to guarantee the `The master can recreate the worker after the worker commits suicide.`, like :https://github.com/hanfei1991/microcosm/blob/c16e9f05bfcbf90ecb6a48be9210b3bfa8bcc5c1/lib/config/timeouts.go#L23

maxshuang: Sometimes we may add other timeout in this critical path which violate the guarantee, like the `workerExitWaitForMasterTimeout` here will increase the worker exit timespan.

maxshuang: Consider such time sequence：
t0: doPoll() -> PollExit() -> get app error -> enter workerStateNormal branch
t1: runWatchDog ->  workerTimeout -> set workerStateExit
t2: overide and set workerState to workerHalfExit
t3: wait workerExitWaitForMasterTimeout(15s) to exit

Because a time slice in Load() and Store()

liuzix: resolved in the latest commit.

maxshuang (approved):

PR #387, opened 2022-05-13
Add Multi-projects/tasks  prometheus metric support

maxshuang:
close issue: #398


pkg/tenant/tenant.go
```diff
@@ -1,11 +1,21 @@
...
+)
 
 // tenant const variables
 const (
 	FrameTenantID       = "dfe_root"
 	TestTenantID        = "dfe_test"
 	DefaultUserTenantID = "def_default_user"
 )
+
+// ProjectInfo is the tenant/project information which is consistent with DBaas
```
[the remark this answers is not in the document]
maxshuang: we don't need much work on tenant/project.
I think maintaining  both `tenantID` and `projectID` is fine.

pkg/promutil/implement_test.go
```diff
@@ -0,0 +1,88 @@
+package promutil
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+)
+
```
amyangfei, line 9: Can we test every `NewXXX` methods of wrappingFactory

pkg/promutil/registry.go
```diff
@@ -0,0 +1,86 @@
...
+	cls, exists = r.collectorByWorker[workerID]
+	if !exists {
+		cls = make([]prometheus.Collector, 0)
+	}
+	cls = append(cls, c)
+	r.collectorByWorker[workerID] = cls
+}
+
+// Unregister unregisters all Collectors of the specified worker
+func (r *Registry) Unregister(workerID libModel.WorkerID) {
```
amyangfei, line 67: We can add ad test to verify `Unregister` can work as expected

pkg/promutil/implement.go
```diff
@@ -0,0 +1,119 @@
...
+	if prefix != "" {
+		opts.Namespace = prefix + "_" + opts.Namespace
+	}
+	cls := opts.ConstLabels
+	for name, value := range constLabels {
+		if _, exists := cls[name]; exists {
+			log.L().Panic("duplicate label name", zap.String("label", name))
+		}
+		cls[name] = value
+	}
```
amyangfei, line 86: We can abstract this logic into a function and reuse it.

amyangfei (commented):
rest lgtm

pkg/promutil/implement.go
maxshuang: solve it in last commit

amyangfei (approved):

Issue #390, opened 2022-05-13
go-mysql-server is hard to use for backend DB


maxshuang:
go-mysql-server  claims to be the 100% mysql-compatible in-memory DB. But it has follow drawbacks:
(1) too many mysql feature not implement
(2) don't support unqiue index yet
(3) concurrent transaction may trigger some data-race in server

amyangfei:
closed by #391

PR #391, opened 2022-05-13
feat(test): support sqlite as backend DB for unit tests

maxshuang:
close issue: #390 

Modify:
(1) change the index name in model to avoid index name conflict in sqlite
(2) replace the go-mysql-server with in_memory sqlite

go-mysql-server claims to be the 100% mysql-compatible in-memory DB. But it has follow drawbacks:
(1) too many mysql feature not implement
(2) don't support unqiue index yet
(3) concurrent transaction may trigger some data-race in server

So we replace the Backend  test DB to sqlite, more things need to be notice:
(1) It has different SQL semantic with MySQL, but with the gorm medium layer, we can use sqlite to some extend.
(2) some bugs like concurrent transaction may fail need to be fixed.


pkg/orm/mock_test.go
```diff
@@ -30,24 +31,25 @@ func TestGenEpochMock(t *testing.T) {
 	}
 	require.Equal(t, int64(11), epoch)
 
-	// FIXME: go-mysql-server concurrent transaction may cause data race
-	// need mutex protection
+	// FIXME: 'no such table: logic_epochs' will raise, why??
 	/*
```
amyangfei, line 34: Can we enable this case now

pkg/orm/mock.go
```diff
@@ -2,146 +2,37 @@ package orm
...
-	"github.com/dolthub/go-mysql-server/sql/information_schema"
-	"github.com/hanfei1991/microcosm/pkg/meta/metaclient"
-	"github.com/hanfei1991/microcosm/pkg/tenant"
-	"github.com/phayes/freeport"
-	"github.com/pingcap/failpoint"
+	cerrors "github.com/hanfei1991/microcosm/pkg/errors"
 	"github.com/pingcap/log"
-	"github.com/pingcap/tiflow/pkg/retry"
+	"go.uber.org/zap"
+	"gorm.io/driver/sqlite" // Sqlite driver based on GGO
```
amyangfei, line 10: What is `GGO` short for, or is it typo of `CGO`?

pkg/orm/mock_test.go
maxshuang: We can't do that now. A new connection(creat when we start a tx) will create a new DB, and cause 'no such table' error consequently.
But if we enable 'cache=shared' option for sqlite, we can solve this problem and cause other problems because we enable 't.Parallel()' and cases will shared the same DB at the mean time.

pkg/orm/mock.go
maxshuang: Already remove the comment.

liuzix (approved):

PR #397, opened 2022-05-18
pkg/notifier: implement notifier

liuzix:
- Implemented a Single-Producer Multiple-Consumer event notifier.

### DIFF
--- a/pkg/containers/slice_queue.go
+++ b/pkg/containers/slice_queue.go
@@ -1,0 +1,109 @@
+package containers
+
+import "sync"
+
+// SliceQueue is a FIFO queue implemented
+// by a Go slice.
+type SliceQueue[T any] struct {
+	mu    sync.Mutex
+	elems []T
+
+	C chan struct{}
+
+	pool *sync.Pool
+}
+
+// NewSliceQueue creates a new SliceQueue.
+func NewSliceQueue[T any]() *SliceQueue[T] {
+	return &SliceQueue[T]{
+		C:    make(chan struct{}, 1),
+		pool: &sync.Pool{},
+	}
+}
+
+func (q *SliceQueue[T]) Add(elem T) {
+	q.mu.Lock()
+
+	signal := false
+	if len(q.elems) == 0 {
+		signal = true
+		if q.elems == nil {
+			q.elems = q.allocateSlice()
+			q.elems = q.elems[:0]
+		}
+	}
+
+	q.elems = append(q.elems, elem)
+	q.mu.Unlock()
+
+	if signal {
+		select {
+		case q.C <- struct{}{}:
+		default:
+		}
+	}
+}
+
+func (q *SliceQueue[T]) Pop() (T, bool) {
+	q.mu.Lock()
+
+	var zero T
+	if len(q.elems) == 0 {
+		q.mu.Unlock()
+		return zero, false
+	}
+
+	ret := q.elems[0]
+	q.elems[0] = zero
+	q.elems = q.elems[1:]
+
+	if len(q.elems) == 0 {
+		q.freeSlice(q.elems)
+		q.elems = nil
+	} else {
+		// non empty queue
+		select {
+		case q.C <- struct{}{}:
+		default:
+		}
+	}
+
+	q.mu.Unlock()
+	return ret, true
+}
+
+func (q *SliceQueue[T]) Peek() (retVal T, ok bool) {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if len(q.elems) == 0 {
+		ok = false
+		return
+	}
+
+	return q.elems[0], true
+}
+
+func (q *SliceQueue[T]) Size() int {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	return len(q.elems)
+}
+
+func (q *SliceQueue[T]) allocateSlice() []T {
+	ptr := q.pool.Get()
+	if ptr == nil {
+		return make([]T, 0, 16)
+	}
+
+	return *(ptr.(*[]T))
+}
+
+func (q *SliceQueue[T]) freeSlice(s []T) {
+	var zero T
+	for idx := range s {
+		s[idx] = zero
+	}
+	q.pool.Put(&s)
+}
--- a/pkg/containers/slice_queue.go
+++ b/pkg/containers/slice_queue.go
@@ -26,7 +26,7 @@ func NewSliceQueue[T any]() *SliceQueue[T] {
 	}
 }
 
-func (q *SliceQueue[T]) Add(elem T) {
+func (q *SliceQueue[T]) Push(elem T) {
 	q.mu.Lock()
 
 	signal := false
--- a/pkg/containers/slice_queue.go
+++ b/pkg/containers/slice_queue.go
@@ -8,6 +8,11 @@ type SliceQueue[T any] struct {
 	mu    sync.Mutex
 	elems []T
 
+	// C is a signal for non-empty queue.
+	// A consumer can select for C and then Pop
+	// as many elements as possible in a for-select
+	// loop.
+	// Refer to an example in TestSliceQueueConcurrentWriteAndRead.
 	C chan struct{}
 
 	pool *sync.Pool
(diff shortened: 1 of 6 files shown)


pkg/notifier/notifier.go
```diff
@@ -0,0 +1,176 @@
...
+func (r *Receiver[T]) close() {
+	r.closed.Store(true)
+	r.closeOnce.Do(
+		func() {
+			close(r.C)
+		})
+}
+
+// Close closes the receiver
+func (r *Receiver[T]) Close() {
```
amyangfei, line 50: What is this function used for? Should the receiver be removed from the notifier list when it is closed/removed?

pkg/containers/slice_queue.go
```diff
@@ -0,0 +1,109 @@
+package containers
```
maxshuang, line 1: What's the motivation to implement such a FIFO queue? 
Do you have some performance tests?

1. Or can we just use channel ? 
2. Or implement the FIFO queue with the list + object pool?

pkg/containers/slice_queue.go
```diff
@@ -0,0 +1,109 @@
...
+
+import "sync"
+
+// SliceQueue is a FIFO queue implemented
+// by a Go slice.
+type SliceQueue[T any] struct {
+	mu    sync.Mutex
+	elems []T
+
+	C chan struct{}
```
maxshuang, line 11: What's this design for?
Why we always need to consume the `C`?

pkg/containers/slice_queue.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	ret := q.elems[0]
+	q.elems[0] = zero
+	q.elems = q.elems[1:]
+
+	if len(q.elems) == 0 {
+		q.freeSlice(q.elems)
+		q.elems = nil
+	} else {
```
maxshuang, line 63: Does it mean that if we don't consume the queue, we will block  in the  queue.Pop even the queue is not empty?
Is it by design? 
It seems that we only want a notifier mechanism to  tell us that queue is not empty.

pkg/notifier/notifier.go
```diff
@@ -0,0 +1,176 @@
...
+					receiver := value.(*Receiver[T])
+
+					if receiver.closed.Load() {
+						return true
+					}
+
+					select {
+					case <-n.closeCh:
+						return false
+					case receiver.C <- event:
```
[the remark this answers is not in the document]
maxshuang: what scenarios this notifier will be used?
Original receiver.C  is 16, will it block the whole run? Or 1024 will be a better one?

pkg/notifier/notifier.go
```diff
@@ -0,0 +1,176 @@
...
+// Notifier is the sending endpoint of a single-producer-multiple-consumer
+// notification mechanism.
+type Notifier[T any] struct {
+	receivers sync.Map // receiverID -> *Receiver[T]
+	nextID    atomic.Int64
+
+	queue *containers.SliceQueue[T]
+
+	closeCh       chan struct{}
+	synchronizeCh chan struct{}
```
maxshuang, line 25: Can you add more comment on the `synchronizeCh`?
Can't the point.

pkg/containers/slice_queue.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return *(ptr.(*[]T))
+}
+
+func (q *SliceQueue[T]) freeSlice(s []T) {
+	var zero T
+	for idx := range s {
+		s[idx] = zero
+	}
```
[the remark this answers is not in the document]
liuzix: Addressed in the latest commit.

pkg/notifier/notifier.go
liuzix: My mistake. Fixed in the latest commit.

pkg/notifier/notifier.go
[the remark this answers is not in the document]
maxshuang: `Single-Producer Multiple-Consumer` has more concret meaning in other language which means msg is consume by one consumer exclusively.

Please make a more concret comment for the notifier.

pkg/containers/slice_queue.go
liuzix: The motivation is:
- A Go channel is a bounded channel. Sometimes we need an unbounded channel, especially when the number of buffered elements can vary widely, or when it is sub-optimal for the producer to be blocked.

pkg/containers/slice_queue.go
```diff
@@ -0,0 +1,109 @@
...
+	signal := false
+	if len(q.elems) == 0 {
+		signal = true
+		if q.elems == nil {
+			q.elems = q.allocateSlice()
+			q.elems = q.elems[:0]
+		}
+	}
+
+	q.elems = append(q.elems, elem)
```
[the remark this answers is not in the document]
liuzix: - We do not have to guarantee fix-length of the slices. It is reasonable that you have some concern here. The size of the slices that we cache in the `pool` should have a limit. I'll add a todo here.
- Implementing two-level array management is too complicated. We can evaluate this in the future.

pkg/containers/slice_queue.go
```diff
@@ -0,0 +1,109 @@
...
+	if ptr == nil {
+		return make([]T, 0, 16)
+	}
+
+	return *(ptr.(*[]T))
+}
+
+func (q *SliceQueue[T]) freeSlice(s []T) {
+	var zero T
+	for idx := range s {
```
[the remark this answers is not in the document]
liuzix: These redundant operations are removed now.

pkg/notifier/notifier.go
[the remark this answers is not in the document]
maxshuang: > A congested receiver.C will not block the producer, which is enough for our purposes.

What scenarios this notifier will be used?
But It may block all receiver. Right? Why is it reasonable?

pkg/notifier/notifier.go
maxshuang: Get the point.
`synchronizeCh` is to achieve async-close the receiver `C` safely.
Pls add some comment if possible. Thanks.

pkg/notifier/notifier.go
```diff
@@ -0,0 +1,168 @@
...
+	n.wg.Wait()
+
+	n.receivers.Range(func(_, value any) bool {
+		receiver := value.(*Receiver[T])
+		receiver.Close()
+		return true
+	})
+}
+
+// Flush flushes all pending notifications.
```
maxshuang, line 117: `Flush` method will wait the queue being queue `Synchronously`.
It has some implicit consumption that producer need to wait this `Flush` too, instead of `letting it happen by chance`.

Pls add more comment on it. Thanks.

pkg/notifier/notifier.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"github.com/pingcap/errors"
+	"go.uber.org/atomic"
+
+	"github.com/hanfei1991/microcosm/pkg/containers"
+)
+
+type receiverID = int64
+
+// Notifier is the sending endpoint of a single-producer-multiple-consumer
```
maxshuang, line 16: // Notifier is the sending endpoint of a single-provider-multiple-publisher

pkg/notifier/notifier.go
```diff
@@ -0,0 +1,186 @@
...
+	queue *containers.SliceQueue[T]
+
+	closed        atomic.Bool
+	closeCh       chan struct{}
+	synchronizeCh chan struct{}
+
+	wg sync.WaitGroup
+}
+
+// Receiver is the receiving endpoint of a single-producer-multiple-consumer
```
maxshuang, line 32:
```suggestion
// Receiver is the receiving endpoint of a single-provider-multiple-publisher

pkg/notifier/notifier.go
maxshuang: Because I has a time slice between `Flush` return and `Register a receive` and it can't guarantee some semantic like `all new msg will be sent to newly added receiver`.

amyangfei (approved):

amyangfei:
ptal @maxshuang

